PR #7, opened 2017-01-21
Create nohog2.c

skaftetryne:
Small accessory that improves multitasking when using the TOS AES under FreeMiNT.

### DIFF
--- a/nohog2/nohog2.c
+++ b/nohog2/nohog2.c
@@ -1,0 +1,95 @@
+/*
+ * no_hog2.acc
+ *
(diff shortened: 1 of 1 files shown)


mikrosk:
Great, you have really done it! Can you extend the patch so it includes being part of the freemint tree? I.e. the patch would be against the `tools` directory, with `nohog2` being included in the top-level Makefile. That way you'd see whether it compiles well on all platforms and after your PR is accepted, it would be automatically there.

skaftetryne:
Can you revert/close this pull request without committing it? Couldn't figure out how to do it myself.

PR #8, opened 2017-01-22
nohog2.acc by Jo Even

mikrosk:
Jo was too impatient to get #7 up and running. Clean and with source code ACC for making GEM=ROM work. @skaftetryne that wasn't that hard, was it.


mikrosk:
Btw Jo has raised a good point -- for tools like this we'd need a minimal startup code because now mintlib eats about 100 KB.

mikrosk:
Hey @skaftetryne, your nice little tool has made it into `master`! :)

PR #11, opened 2017-01-28
enable optional link to libcmini instead of mintlib

mfro0:
Curious if this works:
Dear @freemint/kernel 

meant as a template for (optionally) linking of MiNT tools (that don't need POSIX) against libcmini instead of mintlib.

For the nohog2.acc, this ends up in a reduction of the executable file from about 108 kB (mintlib) to 5268 bytes without loss of functionality.

Just build (from within the nohog2 directory) with

`M68K_ATARI_MINT_CROSS=yes LIBCMINI=yes make`

I also took the freedom to make the Makefile ColdFire-aware:

`M68K_ATARI_MINT_CROSS=yes LIBCMINI=yes CPU=v4e make`

will create a ColdFire executable.

This could also be added to other tools if it became general consensus that this is the way to go.

Get libcmini from AtariForge:

`svn co http://atariforge.org/svn/libcmini`

should be done from the same directory your freemint directory resides in. Don't forget to make libcmini first:

`cd libcmini/libcmini; make`

### DIFF
--- a/CONFIGVARS
+++ b/CONFIGVARS
@@ -48,7 +48,7 @@ LIBCMINI_CPU := m5475
 else
 LIBCMINI_CPU := .
 endif
-LIBCMINI_LIBPATH := $(top_srcdir)/../../libcmini/libcmini/$(LIBCMINI_CPU)
+LIBCMINI_LIBPATH := $(LIBCMINI_PATH)/$(LIBCMINI_CPU)
 LIBCMINI_STARTUP := $(LIBCMINI_PATH)/$(LIBCMINI_CPU)/startup.o
 endif
 
--- a/CONFIGVARS
+++ b/CONFIGVARS
@@ -20,17 +20,13 @@ endif
 ifeq ($(CROSS),yes)
 
 CROSSPREFIX=m68k-atari-mint-
-#CROSSPREFIX=ozk42-
 
 LANG=C
 NATIVECC = gcc
 NATIVECFLAGS = -O -Wall
 CRLF = echo crlf -s
 
-# e.g. when you don't have the mintlib/gemlib/cflib in
-#      the /usr/m68k-atari/mint
 GENERAL = $(M68K_ATARI_MINT_CFLAGS) $(M68K_ATARI_MINT_LDFLAGS)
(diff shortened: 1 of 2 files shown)


mikrosk:
Nice. :) While you are at it, what about enabling it for others, too? fsetter2, hypview, cops, mintloader, xaloader ... they don't seem to require much of the unix part of mintlib.

zabruder:
xaloader is not linked against libc, and all targets where -nostdlib is set, also.

mikrosk:
xaloader (same as mintloader) uses copied files from mintlib (startup etc) so it's even worse.

mfro0:
libcmini uses (more or less verbatim) copies of mintlib startup files as well, but avoids pulling in the timezone stuff. If I remember correctly, this was what made the binaries explode since it basically pulls everything else.

Note that the libcmini link doesn't get checked with travis for now. If that's what we want, there is a little more work to do.

> While you are at it, what about enabling it for others, too?
Certainly, this could be done for other tools as well. I'd rather see some feedback first what others think. I assume not everybody likes external library dependencies.

mfro0:
`xaloader` is 5210 bytes in my installation. Certainly not much to optimize there ;)

mikrosk:
@mfro0 but that's the point, they are so small because they use duplicate files from mintlib. So it's redundant files all over the place. And if someone (like Jo) wants to implement something new (like nohog2.acc), he would be tempted to either use mintlib or copy the stuff from `xaloader` again. Right now, these projects use the same startup files and are great candidates for libcmini:
- xaaes/src.km/xaloader
- sys/usb/src.km/udd/mouse
- sys/usb/src.km/udd/eth
- sys/usb/src.km/udd/storage
- sys/usb/src.km/loader
- sys/usb/src.km
- sys/usb/src.km/ucd/unicorn
- tools/mintload
- tools/usb

atic-atac:
Looks good. I'm sure we can make use of this for some other parts of the build.

mikrosk:
@mfro0 now you have a thumb up from Alan, feel free to propagate libcmini to as many freemint projects as possible ;)

(then I will make sure they are compiled with libcmini on so they are offered in auto builds in the smallest possible size)

PR #14, opened 2017-02-04
Fix building with gemlib 0.44.x

mikrosk:
@vinriviere 's toolchain uses older gemlib so the travis build didn't catch it :-/

### DIFF
--- a/tools/cops/cpx.h
+++ b/tools/cops/cpx.h
@@ -1,6 +1,6 @@
 /*
  * $Id$
- * 
+ *
  * COPS (c) 1995 - 2003 Sven & Wilfried Behne
  *                 2004 F.Naumann & O.Skancke
  *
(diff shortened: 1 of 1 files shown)


vinriviere:
I haven't upgraded GemLib in cross-tools yet because there has been a lot of changes lately, and I feared regressions. If something (i.e. Travis CI?) could confirm that the current Git GemLib is OK, I would me more confident in upgrading the cross-tools with it.

Also, it might be good that Travis CI uses daily builds of the libraries (including GemLib) instead of using the ones provided with the cross-tools.

vinriviere:
Please provide minimal patches in pull requests. Your current patch has white space changes unrelated to the main issue.

mikrosk:
There's one more thing, unless you've upgraded the ubuntu repository lately we're still fetching `trusty` libraries so even if you do upgrade it, we still depend on your decision to "backport" stuff back to `trusty`.

About the blanks, I wasn't sure what is the right thing to do -- providing a separate PR just for blanks seemed silly and on the other hand, it was only few of them so I included it. Maybe some guidelines in wiki would be useful. :-)

vinriviere:
Don't worry, in my next cross-tools update, I will update both yakkety and trusty binaries.

Regarding to the blanks (or any other patch): 1 purpose = 1 patch. Generally, it is bad practice to fix blanks or other stuff in parts of code unrelated to the main topic. Of course, a separate patch to fix *all* whitespace issues (but only that) would make sense. With respect of these rules, the history stays clean, all patches stay readable an easily revertable.

mikrosk:
I admit, I still think in the CVS universe. :-) Of course I could have created two separated commits in the same PR. I guess it's not worth the effort fix this five blanks but for the future I'll be more git oriented for sure. :-)

PR #17, opened 2017-02-04
mintloader's default path is not \\mint but <SYSDIR>.

mikrosk:
This works thanks to the shared source tree and it's much more helpful. The idea is to update/replace only mintload.prg in AUTO and not care about the platform on which the loader is executed. In the future, this opens the door for memory initialisation, MMU setup etc.


mikrosk:
I guess this wont offend anyone as I'm the solely user of this tool but wait a moment and you'll see an asskicking usage of this nice little feature. :-)

Issue #12, opened 2017-01-29
Make build system more consistent


mfro0:
While the kernel and XaAES appear to support all the CPU subtargets (and compile all of them each run into their respective .compile($(CPU)) subdirs even if you usually only want one of them), this is not consistent elsewhere.

Some of the targets honour $(CPU) (and build only for the specified target), others completely ignore the setting and even try to compile incompatible asm and fail.

I'd like to see this changed into a scheme where either everything is build for all supported targets (using the .compile$(CPU) subdir method consistently everywhere) or only for the one specified. I'd prefer the latter. E.g. you would then just do

`M68K_ATARI_MINT_CROSS=yes CPU=ara make`

from the top dir and would get all the binaries for the specified $(CPU) subtarget **only**

mikrosk:
Just for the reference, ideally the whole ecosystem would be like this:
- three platforms: ST (current 000), TT/Falcon/CT2 (030), AB040/Milan/Hades/CT60/Aranym (040,060,mil,ara) with feature autodetection
- make CPU=xxx, where xxx would be any value -mcpu= accepts (i.e. -m68000, m68020-60, m68060, mcpu=5475, ...)

So you could make your Milan build 040 optimised while your friend's CT60 build would be 060 optimised. Not sure about the FireBee, whether it needs a separate platform or not, right now 030/040 separation is needed because of the PMMU so this would imply that the FB needs another one as well.

But this is a future. Right now, having make CPU=xxx working would be more than enough.

mikrosk:
Another strange inconsistency: Milan kernel is compiled for 020-60 (why not 040-60?) but it uses xaaes040.km (compiled for 040). On the other hand, TT/Falcon kernel is compiled for 030 but it uses xaaes030.km compiled for 020-60, WTF?

I know the speed difference is not the problem here (neither is compatibility, as all 020+ builds fall back onto 020-60 libgcc&friends), it's just another symptom of the aged build process.

mikrosk:
@atic-atac are you sure you want this closed? I mean, my recent patches have improved some bits here and there but the ultimate goal, top level `M68K_ATARI_MINT_CROSS=yes CPU=ara make` is still not there (mostly because of the kernel build handling but also don't forget network drivers).

atic-atac:
I didn't close it, github looks like it did based on the merge.

paulwratt:
[...]

PR #18, opened 2017-02-04
Enable `make CPU=xxx` also for tools

mikrosk:
All it took was to move it into the toplevel `RULES`.

Toplevel `make CPU=xxx` doesn't work of course but now you can use the `CPU` variable pretty much anywhere, if you know what are you doing. Doesn't make #12 resolved but it's a little bit closer.

P.S. Surprisingly, many build scripts override `$(CPU)` (like `libkern`) for their needs, although not all of them. So maybe the next step would be to ensure that the `$(CPU)` is honoured where appropriate else override with the correct/desired value.

### DIFF
--- a/RULES
+++ b/RULES
@@ -7,6 +7,12 @@ include $(srcdir)/SRCFILES
 include $(srcdir)/EXTRAFILES
 include $(srcdir)/MISCFILES
 
+ifeq ($(CPU),v4e)
+MODEL = -mcpu=5475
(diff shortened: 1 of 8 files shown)


mikrosk:
Oops, oops, oops. Don't merge yet. My own weapon has got me, build is failing. ;-)

mikrosk:
Should be OK now.

Issue #1, opened 2017-01-19
How to correctly setup deployment


mikrosk:
While Travis CI offers decent tools for testing builds for various flavours, I'm not sure how to take it a step further, to provide compiled binaries for download every time a commit is made.

Ideally, if such builds could be provided by http://freemint.github.io.

Especially nice would be to make it for [tos.hyp project](https://github.com/freemint/tos.hyp) which would be presented as a website (currently at http://toshyp.atari.org).

mikrosk:
It seems travis supports at least tag-based builds for github: https://docs.travis-ci.com/user/deployment/releases/

PR #20, opened 2017-02-11
Reduce the number of CPU/kernel targets

mikrosk:
This may be quite controversial but please hear me out. :-) While working on #1 I've realised how messy the whole build ecosystem is and I don't mean only #12. The user visible list of changes:
- only three CPU targets exist: `000`, `020-60` and `v4e` (same as our gcc multilib configuration) plus one special, `deb`
- `xaaes.km` fully adheres to this setup, there's really no need for anything else (I've merged `000` and `sto` [st only] to one, there's no point of having a `000` target with extended feature set)
- `mint*.prg` can't be fully presented like this, mostly because of the different 030/040 MMU setup and the fact that Aranym and Milan are not detected on the fly. So I've removed only 040 target (useless) and merged it with the 060 target and removed the 020 target (again, useless); all kernels respect the new CPU setup
- the idea is, from now on (and when #1 is done -- I've basically have it up and running), to provide three separate archives, (yes, you've guessed it) based on the CPU class: ST/E, TT/Falcon, Milan/CT60/Aranym ... no more mess with 020, 030, 020-60, 040, 060 archives, for christ's sake, sometimes even I as a developer don't know what I'm supposed to run on my computer

But this is not all, with the new `mintloader` everything would be done on the fly! So you know you have a Milan, fine, download milan_hades_ct60_aranym_mint.zip (yeah, we need to work out on the name ;)), unpack it and ... that's it! No more messing with the right modules. Even more visible it is for the FireBee where's a real danger to run modules for an incompatible CPU.

XaAES doesn't need to provide all the crazy `km` files, there's one `km` per archive, more than enough.

Also, don't worry about optimisation. I'm quite ashamed to admit it but it was me who had introduced `M68060` to the kernel 10 years ago. :-) Looking at my enthusiasm now I must say it wasn't very useful. `-m68020-60` is basically the same as `-m68020` except the fact that some exotic instructions like 64/32 -> 32 or 64*32 -> 64 are replaced with a gcc subroutine, i.e. something we don't even use in the kernel.

And last but not least, compilation time is now *much* lower because `libkern` and friends has to be compiled only for the three (four) CPU targets.

I have a few more tricks in my sleeve for auto builds to make user testing/deployment as smooth as possible (stay tuned) but it needs this to make it work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ compiler:
    - m68k-atari-mint-gcc
 
 before_install:
+    - echo $TRAVIS_PULL_REQUEST_SLUG
+    - echo $TRAVIS_REPO_SLUG
   - 'sh ./.install-cross-mint.sh'
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ compiler:
    - m68k-atari-mint-gcc
 
 before_install:
-    - echo $TRAVIS_PULL_REQUEST_SLUG
-    - echo $TRAVIS_REPO_SLUG
+  - echo ${TRAVIS_PULL_REQUEST_SLUG}
+  - echo ${TRAVIS_REPO_SLUG}
   - 'sh ./.install-cross-mint.sh'
 
 script:
--- a/sys/KERNELDEFS
+++ b/sys/KERNELDEFS
@@ -16,7 +16,6 @@
 # -DM68000		build for 68000 (no mmu, no caches)
 # -DM68030		build for 68030
 # -DM68040		build for 68040-060
-# -DM68060		build for 68060
 # -DMILAN		build Milan kernel
 # -DARANYM		build Aranym kernel
 
@@ -59,6 +58,13 @@ KERNELDEFAULTDEFS += -DSOFT_UNITABLE
 KERNELDEFAULTDEFS += -DBUILTIN_SHELL
 endif
 
+ifeq ($(kernel),deb)
+MINT = mintdeb.prg
+CPU  = 020-60
+KERNELDEFS = -DDEBUG_INFO -DM68040
+endif
+
+# ST, STE, 020+ MMU-less machines
 ifeq ($(kernel),000)
 MINT = mint000.prg
 CPU  = 000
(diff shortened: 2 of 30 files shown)


vinriviere:
Great! This was the thing to do.

I had a similar idea, but with different groups. The idea was to build complete archives based on CPU: 000, 030, 040, 060, v4e. So one would just have to pickup the right archive based on his CPU, regardless of other hardware. To achieve this, Milan and ARAnyM should be detected, and the functionality of their respective kernels should be merged into mint040.prg.

This also opens the question of support of [68030 without FPU in GCC](https://github.com/freemint/m68k-atari-mint-gcc/issues/1).

mikrosk:
@vinriviere yeah, I wish we had one, universal tree and one could choose whatever CPU he likes but unfortunately, we don't. So I wanted to do a least _something_ to make snapshots/releases easier until someone braves enough dives in into Milan/Aranym incompatibilities.

atic-atac:
I'm not a fan of this because folks will always want optimized code for their CPU. So I'm inclined to leave these targets.

It would be nice for Milan and ARAnyM to be detected, but the other thing with these builds is memory consumption. I know we have different builds for 000 CPUs just to cater for folks with limited RAM.

mikrosk:
@atic-atac have you heard anyone with a 040+ machine ever to complain about memory consumption?

I made the 000 target "ST ONLY" (i.e. the lowest possible memory usage), as I don't see a point to do it otherwise.

And about the "folks", well... there'a always an option to override the CPU variable. Believe me I so do know how they feel but it's just pointless, there's absolutely no reason to replace `-m68020-60` with `-m68030` or even `-m68020-60` with `-m68060`. Not. A. Single. One. It's only about the feeling "I have it compiled *just* for my machine". :)

Feel free to postpone this one until I present you my branch with the autobuilds, it will become much more obvious how this PR greatly simplifies testing and deployment.

atic-atac:
Present people options, dont limit their choices.

mikrosk:
So how many flavours the Linux kernel offers in your distro? ;-)

But I see your point.

Ideally, #12 would ultimately solve this but that's too overwhelming for me, I wanted to start with something simpler. Anyway, as I say, maybe you and the folks ;) will change your opinion later.

atic-atac:
For the kernel most users, they just want to grab a binary and install, sending them to git to go and build opens up a whole world of pain for them. That's effectively what you are telling people to do if they want a variation on what's pre-compiled. The kernel is such a small part of the system, so we should be giving choices and pre-compiled, documented features for each. Not limiting them.

mikrosk:
@atic-atac but so far you've been very generic. So what exactly your concern is? It can't be performance, memory consumption isn't increased at all in this PR, so are we really talking only about the nice feeling "-m68CPU for my CPU" ? How come you're not concerned about the terrifying limits mintlib poses on apps (see https://github.com/freemint/m68k-atari-mint-gcc/issues/1) ... would you accept a patch for every possible m68k CPU+softfloat/hardfloat ?

atic-atac:
Don't extend the conversation beyond what's here. You start talking about my distro, then mintlib. Jeesh.

Yes, I'm talking about the nice feeling of having an 040 CPU kernel compilation for the 040 machine, and the 030 CPU kernel for the TT. etc etc etc etc.

mikrosk:
I just wanted to see your point of view in general, applied on the Linux world, mintlib etc, I meant no disrespect.

mikrosk:
Rebased on #27

mikrosk:
Oops, it seems I can't recover a branch from PR after I'd deleted it. I've recreated a (rebased) branch in https://github.com/mikrosk/freemint/tree/pr-20 however it doesn't seem it would get accepted any time soon so I'm not going to create another PR for it.

PR #31, opened 2017-04-05
Add Svethlana driver

jvaltane:
This adds Svethlana driver and sources needs some polishing.

I created new pull request because old v0.1 driver wanted to as base of the driver. Also I made an mistake using forks master, own branch is better. I compiled and tested that it worked with SuperVidel FW9. Then replaces v0.1 sources with v0.9 sources.

I think next steeps would be to remove dead code and polish debug stuff.


mikrosk:
Just for the record, you don't need to create new PR when rebasing -- just rewrite the pushed history. It's your branch, you can do force pushes as much as you please.

Anyway, looks good to me, thank you for taking time to do this!

PR #34, opened 2017-05-07
Fix 'Open console at output'

mikrosk:
Now it works without logging enabled, too. Reported on the mailing list.


vinriviere:
The last build works perfectly. Many thanks to @mikrosk 
Please merge!

PR #36, opened 2017-05-14
Add Apollo 68080 detection

vinriviere:
This forces the _CPU coookie to 68040, as this processor has the most
complete instruction set. This is recommended by the Apollo Team.

### DIFF
--- a/sys/arch/detect.S
+++ b/sys/arch/detect.S
@@ -281,10 +281,21 @@ no_040:	lea	exit.w(PC),a0	// change vectors back to 'exit'
 
 	bra.s	exit
 
-// 68040 or 68060.
+// 68040 or higher.
 
 x040:	moveq	#40,d0		// assume 68040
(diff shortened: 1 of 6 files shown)


vinriviere:
Previously, the Apollo 68080 was detected as 68060. Not strictly wrong, as the Apollo 68080 has all features of 68060 (except MMU and currently FPU). But the Apollo Team prefers 68040 detection as the full instruction set is implemented such as movep, etc.

mikrosk:
How the Apollo behaves when executing PMMU instructions then? I'm wondering whether `no_mem_prot` shouldn't be set when the Apollo is detected?

vinriviere:
It works well because MMU is detected as missing, as expected.

vinriviere:
Great!  :thumbsup:
Amiga Vampire users will now be able to enjoy PmDoom in GEM window.

PR #38, opened 2017-05-19
Cosmetic changes to keyboard tables

mikrosk:
- added updated files by Dima
- added uk2us.tbl by Lonny (only *.src was available)
- shortened some tbl filenames to make them more TOS friendly
- removed src files from `mktbl` share folder (they are not really needed for a *binary* distribution + their filenames are very TOS-unfriendly)


mikrosk:
Rebased against master (so I get the .ttp files in /bin)

Issue #40, opened 2017-05-22
Forced fsck on every reboot


mikrosk:
Just to write this down due to its popularity -- this isn't a kernel issue. I've done a few experiments -- the issue depends on `e2fsprogs` used. By default, SpareMiNT had used [e2fsprogs 1.27](https://freemint.github.io/sparemint/sparemint/html/packages/e2fsprogs.html) and this binary got into EasyMiNT (1.70, 1.8x) as well.

Then M.A.Kehr got back to EasyMiNT (1.9x) and with the aim to make a FireBee compatible installer took Vincent's newer [e2fsprogs 1.42.12](http://vincent.riviere.free.fr/soft/m68k-atari-mint/archives/mint/e2fsprogs/) and this is where everyone's troubles start. For some reason `fsck.ext2` from 1.42.12 doesn't set the clean flag properly. What's worse, after the partition has been created this way, the older package's tools can't be used anymore.

I guess @vinriviere is way too busy to investigate right now so until this is fixed I'd recommend to ignore EasyMiNT 1.9x (if one wishes to use EasyMiNT) and stick with the older one. Or at least replace `c:\easymint\mkfs.ttp` with `mke2fs` from 1.27 package.

This is of course bad news not only for everyone who can't afford to reformat ext2 partition but also to FireBee users which are stuck without ColdFire port of e2fsprogs 1.27.

th-otto:
Maybe thats another problem, but i don't think that this is the cause of fsck running a full check. The clean flag should be set by the kernel when unmounting the device. If fsck runs a check, that means that this flag was not set by the kernel. Or newer versions of fsck expect something else to be set/unset, but this could be fixed in mint.

mikrosk:
@th-otto I tried latest kernel with older fsck.ext2 and it works OK. So kernel is setting whatever is supposed to set, the newer fsck.ext2 just don't understand it. But you're right, maybe some public API/struct has changed upstream and FreeMiNT just doesn't reflect that.

vinriviere:
The most obvious cause of trouble would be I forgot something important regarding to the dirty flag in [the e2fsprogs-1.42 patch](http://vincent.riviere.free.fr/soft/m68k-atari-mint/archives/mint/e2fsprogs/). I don't plan to look at that, but feel free to do so.

vinriviere:
[...]

PR #42, opened 2017-05-29
Don't flush invalid filesystem state during dskchng()

mikrosk:
As a result of discussion in #40 I've come up with following solution. It's quite simple and straightforward. To summarise the situation:

- the `XHDOSLimits()` calls in `XHDI_init()` internally register a media change for all partitions in HD Driver
- when a file system driver is loaded, its mounting state is noted (i.e. whether it has been cleanly mounted; based on its file system state flag), then its file system state flag is set to **invalid** and then it automatically calls `dskchng()` to verify whether a `Mediach()` wouldn't return non-zero code (i.e. that the drive has changed)
- since HD Driver has set all partitions as "definitely changed", every file system driver supporting (and handling) `FS_OWN_MEDIACHANGE` (currently: `fatfs`, `ext2fs` and `minixfs`) flushes its cache and remounts the partition
- trouble is that the file system drivers flush also the state flag which is set to invalid, making the whole file system invalid by next remount (which happens immediately after handling `dskchng()`)

The idea behind fix is that we note whether something has been *written* to given device. If not, every file system driver resets the state flag to **valid** and *then* flushes the cache. Please note this happens only and only if the file system has been mounted cleanly, so the file system driver wont reset an invalid state to valid.

Another approach would be to note whether we're calling `dskchng()` in a "proper" situation (i.e. before any user process could access the partition) and reset the flag based on some function parameter to `dskchng()` (for instance an extended `mode`) but since there's a zillion places where `dskchng()` is called from to me it seems much more complex and error prone.

### DIFF
--- a/sys/block_IO.c
+++ b/sys/block_IO.c
@@ -2422,7 +2422,7 @@ bio_invalidate (DI *di)
 
 	if (di->lock > 1)
 	{
-		BIO_FORCE (("block_IO [%c]: invalidate on LOCKED di", di->drv+'A'));
+		BIO_DEBUG (("block_IO [%c]: invalidate on LOCKED di", di->drv+'A'));
 	}
 
 restart:
@@ -2447,7 +2447,7 @@ restart:
 				u->dirty = 0;
 
 				/* inform user */
-				BIO_ALERT (("block_IO [%c]: bio_invalidate: cache unit not written back (%li, %li)!", di->drv+'A', u->sector, u->size));
+				BIO_DEBUG (("block_IO [%c]: bio_invalidate: cache unit not written back (%li, %li)!", di->drv+'A', u->sector, u->size));
 			}
 
 			/* remove from table */
--- a/sys/block_IO.c
+++ b/sys/block_IO.c
@@ -752,6 +752,7 @@ bio_writeout (DI *di, const void *buffer, ulong size, ulong sector)
 	else
 	{
 		BIO_DEBUG (("bio_writeout [sector %lu]: %lu bytes written back!", sector, size));
+		di->mode |= BIO_MODIFIED;
 	}
 
 	return r;
@@ -1605,7 +1606,7 @@ bio_init_di (DI *di)
 
 	di->major = 0;
 	di->minor = 0;
-	di->mode &= ~(BIO_REMOVABLE | BIO_LOCKABLE | BIO_LRECNO);
+	di->mode &= ~(BIO_REMOVABLE | BIO_LOCKABLE | BIO_LRECNO | BIO_MODIFIED);
 
 	di->start = 0;
 	di->size = 0;
--- a/sys/block_IO.c
+++ b/sys/block_IO.c
@@ -302,8 +302,6 @@ struct cbl
 
 static long	_cdecl rwabs_log	(DI *di, ushort rw, void *buf, ulong size, ulong rec);
 static long	_cdecl rwabs_log_lrec	(DI *di, ushort rw, void *buf, ulong size, ulong rec);
-/*static long	_cdecl rwabs_phy	(DI *di, ushort rw, void *buf, ulong size, ulong rec);*/
-/*static long	_cdecl rwabs_phy_lrec	(DI *di, ushort rw, void *buf, ulong size, ulong rec);*/
 static long	_cdecl rwabs_xhdi	(DI *di, ushort rw, void *buf, ulong size, ulong rec);
 
 
@@ -451,80 +449,6 @@ rwabs_log_lrec (DI *di, ushort rw, void *buf, ulong size, ulong rec)
 
 	return sys_b_rwabs (rw, buf, n, -1, di->drv, recno);
 }
-
-# if 0
-static long _cdecl
-rwabs_phy (DI *di, ushort rw, void *buf, ulong size, ulong rec)
-{
-	register ulong n;
-	register ulong recno;
-
-	add_blkdev_randomness (di->drv);
-
-//	if (rw && (di->mode & BIO_WP_MODE))
-//	{
-//		BIO_ALERT (("block_IO [%c]: attempting to write on a write protected device (block %ld)!", di->drv+'A', (rec << di->lshift)));
-//		return EROFS;
-//	}
-
-	n = size >> di->pshift;
-	recno = rec << di->lshift;
-
-	if (!n || n > 65535UL)
-	{
-		BIO_ALERT (("block_IO [%c]: rwabs_phy: n outside range (%li)", di->drv+'A', n));
-		return ESECTOR;
-	}
-
-# if 0
-	if ((recno + n) > di->size)
-	{
-		BIO_ALERT (("block_IO [%c]: rwabs_phy: access outside partition", di->drv));
-		return ESECTOR;
-	}
-# endif
-
-	BIO_FORCE (("block_IO [%i]: rw = %i, start = %li, recno = %li, size = %li, n = %li", di->major, rw, di->start, recno, size, n));
-	return sys_b_rwabs (rw | 8, buf, n, (recno + di->start), di->major, 0L);
-}
-# endif
-
-# if 0
-static long _cdecl
-rwabs_phy_lrec (DI *di, ushort rw, void *buf, ulong size, ulong rec)
-{
-	register ulong n;
-	register ulong recno;
-
-	add_blkdev_randomness (di->drv);
-
-//	if (rw && (di->mode & BIO_WP_MODE))
-//	{
-//		BIO_ALERT (("block_IO [%c]: attempting to write on a write protected device (block %ld)!", di->drv+'A', (rec << di->lshift)));
-//		return EROFS;
-//	}
-
-	n = size >> di->pshift;
-	recno = rec << di->lshift;
-
-	if (!n || n > 65535UL)
-	{
-		BIO_ALERT (("block_IO [%c]: rwabs_phy_lrec: n outside range (%li)", di->drv+'A', n));
-		return ESECTOR;
-	}
-
-# if 0
-	if ((recno + n) > di->size)
-	{
-		BIO_ALERT (("block_IO [%c]: rwabs_phy_lrec: access outside partition", di->drv));
-		return ESECTOR;
-	}
-# endif
-
-	BIO_FORCE (("block_IO [%i]: rw = %i, start = %li, recno = %li, size = %li, n = %li", di->major, rw, di->start, recno, size, n));
-	return sys_b_rwabs (rw | 8, buf, n, -1, di->major, (recno + di->start));
-}
-# endif
 
 static long _cdecl
 rwabs_xhdi (DI *di, ushort rw, void *buf, ulong size, ulong rec)
--- a/sys/block_IO.c
+++ b/sys/block_IO.c
@@ -752,7 +752,6 @@ bio_writeout (DI *di, const void *buffer, ulong size, ulong sector)
 	else
 	{
 		BIO_DEBUG (("bio_writeout [sector %lu]: %lu bytes written back!", sector, size));
-		di->mode |= BIO_MODIFIED;
 	}
 
 	return r;
@@ -1606,7 +1605,7 @@ bio_init_di (DI *di)
 
 	di->major = 0;
 	di->minor = 0;
-	di->mode &= ~(BIO_REMOVABLE | BIO_LOCKABLE | BIO_LRECNO | BIO_MODIFIED);
+	di->mode &= ~(BIO_REMOVABLE | BIO_LOCKABLE | BIO_LRECNO);
 
 	di->start = 0;
 	di->size = 0;
--- a/sys/fatfs.c
+++ b/sys/fatfs.c
@@ -7691,9 +7691,6 @@ fatfs_dskchng (int drv, int mode)
 
 	FAT_DEBUG (("fatfs_dskchng: invalidate drv (change = %li)", change));
 
-	/* I hope this isn't a failure */
-	bio.sync_drv (DI (drv));
-
 	/* invalid all cookies */
 	{
 		register long i;
@@ -7771,7 +7768,33 @@ fatfs_unmount (int drv)
 	if (CLEAN (drv))
 		clean_flag (drv, CLEANFLAG_SET);
 
-	fatfs_dskchng (drv, 1);
+	/* I hope this isn't a failure */
+	bio.sync_drv (DI (drv));
+
+	/* invalid all cookies */
+	{
+		register long i;
+		for (i = 0; i < COOKIE_CACHE; i++)
+		{
+			register COOKIE *c = &(cookies[i]);
+			if (c->dev == drv)
+			{
+				if (c->name)
+				{
+					c_del_cookie (c);
+				}
+			}
+		}
+	}
+
+	/* free the DI (also invalidate cache units) */
+	bio.free_di (DI (drv));
+
+	/* invalidate the BPB */
+	BPBVALID (drv) = INVALID;
+
+	/* free the dynamically allocated memory */
+	kfree (BPB (drv)); BPB (drv) = NULL;
 
 	return E_OK;
 }
--- a/sys/fatfs.c
+++ b/sys/fatfs.c
@@ -7691,6 +7691,15 @@ fatfs_dskchng (int drv, int mode)
 
 	FAT_DEBUG (("fatfs_dskchng: invalidate drv (change = %li)", change));
 
+	if (CLEAN (drv))
+	{
+		if (!BIO_MODIFIED_CHECK(DI (drv)))
+		{
+			FAT_DEBUG (("fatfs_dskchng [%c]: partition is not modified, marking as valid", drv+'A'));
+			clean_flag (drv, CLEANFLAG_SET);
+		}
+	}
+
 	/* I hope this isn't a failure */
 	bio.sync_drv (DI (drv));
 
--- a/sys/fatfs.c
+++ b/sys/fatfs.c
@@ -7691,15 +7691,6 @@ fatfs_dskchng (int drv, int mode)
 
 	FAT_DEBUG (("fatfs_dskchng: invalidate drv (change = %li)", change));
 
-	if (CLEAN (drv))
-	{
-		if (!BIO_MODIFIED_CHECK(DI (drv)))
-		{
-			FAT_DEBUG (("fatfs_dskchng [%c]: partition is not modified, marking as valid", drv+'A'));
-			clean_flag (drv, CLEANFLAG_SET);
-		}
-	}
-
 	/* I hope this isn't a failure */
 	bio.sync_drv (DI (drv));
 
--- a/sys/mint/block_IO.h
+++ b/sys/mint/block_IO.h
@@ -71,9 +71,10 @@ struct di
 # define BIO_WB_MODE		0x02	/* write-back bit (soft) */
 # define BIO_REMOVABLE		0x04	/* removable media */
 # define BIO_LOCKABLE		0x08	/* lockable media */
-# define BIO_LRECNO		0x10	/* lrecno supported */
+# define BIO_LRECNO			0x10	/* lrecno supported */
 # define BIO_WP_HARD		0x20	/* write-protected partition */
 # define BIO_ENCRYPTED		0x40	/* partition is encrypted */
+# define BIO_MODIFIED		0x80	/* partition is modified */
 	
 	ulong	start;			/* physical start sector */
 	ulong	size;			/* physical sectors */
@@ -202,7 +203,7 @@ struct bio
 
 
 /*
- * useful makros
+ * useful macros
  */
 
 # define BIO_RWABS(di, rw, buf, size, recno) \
@@ -212,6 +213,7 @@ struct bio
 # define BIO_WP_CHECK(di)	((di->mode & BIO_WP_MODE) ? ENABLE : DISABLE)
 # define BIO_WB_CHECK(di)	((di->mode & BIO_WB_MODE) ? ENABLE : DISABLE)
 # define BIO_LR_CHECK(di)	((di->mode & BIO_LRECNO) ? ENABLE : DISABLE)
+# define BIO_MODIFIED_CHECK(di) ((di->mode & BIO_MODIFIED) ? ENABLE : DISABLE)
 
 INLINE void
 bio_MARK_MODIFIED (BIO *bio, UNIT *u)
--- a/sys/mint/block_IO.h
+++ b/sys/mint/block_IO.h
@@ -71,10 +71,9 @@ struct di
 # define BIO_WB_MODE		0x02	/* write-back bit (soft) */
 # define BIO_REMOVABLE		0x04	/* removable media */
 # define BIO_LOCKABLE		0x08	/* lockable media */
-# define BIO_LRECNO			0x10	/* lrecno supported */
+# define BIO_LRECNO		0x10	/* lrecno supported */
 # define BIO_WP_HARD		0x20	/* write-protected partition */
 # define BIO_ENCRYPTED		0x40	/* partition is encrypted */
-# define BIO_MODIFIED		0x80	/* partition is modified */
 	
 	ulong	start;			/* physical start sector */
 	ulong	size;			/* physical sectors */
@@ -203,7 +202,7 @@ struct bio
 
 
 /*
- * useful macros
+ * useful makros
  */
 
 # define BIO_RWABS(di, rw, buf, size, recno) \
@@ -213,7 +212,6 @@ struct bio
 # define BIO_WP_CHECK(di)	((di->mode & BIO_WP_MODE) ? ENABLE : DISABLE)
 # define BIO_WB_CHECK(di)	((di->mode & BIO_WB_MODE) ? ENABLE : DISABLE)
 # define BIO_LR_CHECK(di)	((di->mode & BIO_LRECNO) ? ENABLE : DISABLE)
-# define BIO_MODIFIED_CHECK(di) ((di->mode & BIO_MODIFIED) ? ENABLE : DISABLE)
 
 INLINE void
 bio_MARK_MODIFIED (BIO *bio, UNIT *u)
--- a/sys/xfs/ext2fs/ext2sys.c
+++ b/sys/xfs/ext2fs/ext2sys.c
@@ -2150,16 +2150,6 @@ e_dskchng (int drv, int mode)
 
 	DEBUG (("Ext2-FS [%c]: e_dskchng (mode = %i): invalidate drv (change = %li, memory = %li)", drv+'A', mode, change, memory));
 
-	if (!(s->s_flags & MS_RDONLY) && !(s->s_flags & S_NOT_CLEAN_MOUNTED))
-	{
-		if (!BIO_MODIFIED_CHECK(s->di))
-		{
-			DEBUG (("Ext2-FS [%c]: partition is not modified, marking as valid", drv+'A'));
-			s->sbi.s_sb->s_state = cpu2le16 (le2cpu16 (s->sbi.s_sb->s_state) | EXT2_VALID_FS);
-			bio_MARK_MODIFIED (&bio, s->sbi.s_sb_unit);
-		}
-	}
-
 	/* sync the inode cache */
 	sync_cookies ();
 
--- a/sys/xfs/ext2fs/ext2sys.c
+++ b/sys/xfs/ext2fs/ext2sys.c
@@ -2150,6 +2150,16 @@ e_dskchng (int drv, int mode)
 
 	DEBUG (("Ext2-FS [%c]: e_dskchng (mode = %i): invalidate drv (change = %li, memory = %li)", drv+'A', mode, change, memory));
 
+	if (!(s->s_flags & MS_RDONLY) && !(s->s_flags & S_NOT_CLEAN_MOUNTED))
+	{
+		if (!BIO_MODIFIED_CHECK(s->di))
+		{
+			DEBUG (("Ext2-FS [%c]: partition is not modified, marking as valid", drv+'A'));
+			s->sbi.s_sb->s_state = cpu2le16 (le2cpu16 (s->sbi.s_sb->s_state) | EXT2_VALID_FS);
+			bio_MARK_MODIFIED (&bio, s->sbi.s_sb_unit);
+		}
+	}
+
 	/* sync the inode cache */
 	sync_cookies ();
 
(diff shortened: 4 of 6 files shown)


mikrosk:
I guess it's quite obvious but just to be sure: to see this fix really working you have to set the clean/valid flag on each filesystem first (via `dosfsck` and `ext2fsck`). Tested on my CT60+TOS and Aranym+TOS/EmuTOS.

th-otto:
You can also use debugfs on the host to query/set the filesystem state field:

If the image is byte-swapped, convert it first:

dd if=/path/to/diskimage of=tmp conv=swab

To query the state:

state=`debugfs tmp -R "show_super_stats -h" | grep "Filesystem state"`
echo $state

(the backticks are only necessary because debugfs otherwise uses its internal pager to display the results)

To set the state to clean:
debugfs -w tmp -R "set_super_value state 1"

To set it to "not clean":
debugfs -w tmp -R "set_super_value state 0"

After modification, the temp file has to be converted back:

dd if=tmp of=/path/to/diskimage conv=swab

Note: the filesystem image needs to be swapped on the host, if the "Byteswap" option in Aranym is *NOT* set.

mikrosk:
Thanks for the hint, @th-otto ! `debugfs` is even part of the `e2fsprogs` package in SpareMiNT so it's even simpler.

Only complication is fat file system, one has to install [dosfstools](https://freemint.github.io/sparemint/sparemint/html/packages/dosfstools.html) first as it's not part of EasyMiNT.

faucon2001:
Thanks Miro, build 1-19-433 works great on my CT60 Falcon and solved this issue : EasyAraMint is now fully functional on the CT60 :-)  
I now have to dig my Firebee out of the boxes and test it.

vinriviere:
This looks like a good fix at first (as it fixes users trouble), but...

If I understand well, `XHDOSLimits()` triggers a media change event inside HDDRIVER. That event is catched by FreeMiNT's filesystem drivers which in result... flushes their disk cache, really?
Note that any cache has 2 basic operations:
- flush: write any cached data to actual disk
- invalidate: discard any cached data, then reload it from disk when needed

Of course it makes sense to invalidate the cache on media change, since the actual disk is different. But **flushing** the disk would be a total nonsense: it would mean writing old cached data to a potentially different disk. I hope I missed something, and it does not actually work like this...

Second point: Am I wrong, or does your fix only delays the problem? It avoids to flush the disk on that disk change event, but nothing will prevent the filesystem to flush the cache at any time (i.e. when the users runs the `sync` command).

And last remark: why does the problem not show with EmuTOS? Maybe there is something unimplemented in its internal XHDI driver?

atic-atac:
I think the cure here is to run e2fsck.ttp on the partition before FreeMiNT boots at all. That way the filesystem hasn't been mounted by FreeMiNT's filesystem drivers.

Given the hard disk driver automatically "mounts" the drive when it boots, it's always visible as a partition letter anyway.

Unless FreeMiNT is told to mount the filesystem read-only, then later after e2fsck, we remount read-write, which is a much more extensive change, I don't think this change is right.

th-otto:
@atic-atac running e2fsck before mint will be difficult. You could only do that from the auto folder, where you can't pass any arguments like -a. Also before mint, e2fsck may not be able to determine which partitions to check.

@vinriviere I think the problems does not show with EmuTOS, because then HDDRIVER isn't involved. Thats does not necessarily mean its a bug in HDDRIVER, but rather a bad interaction between MiNT/HDDRIVER.

atic-atac:
@th-otto that may be so, but I guess a decent boot loader will allow you to pass arguments, and also e2fsck needs you to specify each partition anyway as it can't determine which partitions to check even in FreeMiNT.

mikrosk:
Thanks for feedback, guys.

@vinriviere : it does happen in EmuTOS, if IDE emulation + HD Driver is used (what makes sense since HD Driver is the source of media change bits, not the OS)

> If I understand well, XHDOSLimits() triggers a media change event inside HDDRIVER. That event is catched by FreeMiNT's filesystem drivers which in result... flushes their disk cache, really?

Really. First writes the cache to disk with `sync_drv()`, *then* invalidates the cache with `free_di()`.

> But flushing the disk would be a total nonsense: it would mean writing old cached data to a potentially different disk. I hope I missed something, and it does not actually work like this...

It didn't make much sense to me either, but it was written by Frank so I assumed he knew what he was doing and I just worked around the bug.

> It avoids to flush the disk on that disk change event, but nothing will prevent the filesystem to flush the cache at any time (i.e. when the users runs the sync command).

My code doesn't avoid any flushing. It only says that *if* something triggers the flush, it resets the dirty/valid flag if nothing has been written to disk yet. Nothing else, everything (most importantly the dirty/valid flag) is flushed/written as before. So it doesn't matter what triggers the flush.

@atic-atac : you're forgetting other filesystems. IMO telling the users that "hey guys, you need to configure your boot manager to run fsck for fat, minix, ext2fs and any possible filesystem in the future" doesn't seem like a solution either. Plus it's awfully anti-Unix, it's like telling Linux users to check their drives from MS-DOS. ;)

Mounting the partitions as read-only sounds much better to me but then again, it's not only about ext2, FreeMiNT not only mounts boot drive (to read mint.cnf etc) but it also *writes* to it (mint.ini, logs etc).

vinriviere:
> My code doesn't avoid any flushing. It only says that if something triggers the flush, it resets the dirty/valid flag if nothing has been written to disk yet.

Ah, in that case your patch looks good, in any case.

I still wonder why EmuTOS (I mean the internal XHDI driver part) does not behave like HDDRIVER, maybe a lack of feature?

skaftetryne:
> that may be so, but I guess a decent boot loader will allow you to pass arguments, and also e2fsck needs you to specify each partition anyway as it can't determine which partitions to check even in FreeMiNT.

Correct, fsck doesn't probe the drives/partitions itself. I have written a small tool that checks each partition/drive and calls the appropriate fsck. It has never occured to me to test this under TOS, but as long as fsck itself does not depend on MiNT being present I see no reason why this should not work.

mikrosk:
@vinriviere

> Ah, in that case your patch looks good, in any case.

That was my intention -- until The Right Fix is found, we at least don't annoy the users (and don't need to revert that original, apart from the regression very useful, patch). However Alan has planted curiosity in my mind about that read only mounting first, doesn't Linux do that as well?

> I still wonder why EmuTOS (I mean the internal XHDI driver part) does not behave like HDDRIVER, maybe a lack of feature?

Looking at: https://github.com/emutos/emutos/blob/master/bios/xhdi.c#L249 I'd say yes, not implemented. :)

atic-atac:
@th-otto it's a temporary solution. The real fix is to make FreeMiNT mount read-only, until such a time in the boot process we can re-mount read/write. 

The fix here is not right at all.

atic-atac:
@mikrosk yes, all unixes mount read-only during boot until later they re-mount read/write.

skaftetryne:
@atic-atac 

> The real fix is to make FreeMiNT mount read-only, until such a time in the boot process we can re-mount read/write.

This is not the real fix. This is exactly the kind of stuff that any modern unix distribution attempts to hide from the user. How do you suggest to handle this in MiNT? How to handle a read-only boot-partition? Or FAT-partitions in general?

@mikrosk 

> yes, all unixes mount read-only during boot until later they re-mount read/write.

But MiNT is not more unix than it is DOS...

mikrosk:
@atic-atac : actually now I see this is not always the case. For instance my Arch Linux runs `fsck` from the `initramfs` stage.

Anyway, I'd say this whole thing has raised following points/questions:

- what should be the *immediate* fix? I hope we all agree the current state of things is unacceptable (fsck every time, alerts on all fat and ext2 partitions)
  - reverting the original patch? (I could live with that, don't need more than 26 drives, some people do)
  - accepting this patch? (not really solving the problem, just working around it)
  - forcing people to run `dosfsck`/`mfsck`/`e2fsck` from AUTO and never after that? (if it's even possible)
  - change the suspicious behaviour as pointed out by @vinriviere (why we're flushing anything if the **drive** has changed?)
- what is The Right Fix? So far we have:
  - read all partitions are read-only; as pointed out this is problematic at least for the FAT partitions the kernels uses for writing stuff (mint.ini and logs)
  - approach `initrd`/`initramfs` behaviour, for instance putting something similar into `mintloader` -- so it would initialise hardware, set up a preliminary root fs, checked all drives, mounted the real one -- since we don't have all that crap with RAIDs and network drives it simplifies things a lot

atic-atac:
@th-otto all partitions that your hard disk driver knows about will be mounted read-write at boot.  FreeMiNT should remount them read-only when it starts, execute the startup scripts as any standard unix system would do, and run fsck during those scripts, and which point the scripts then mount read/write and continue.

There isn't anything sinister here, it's just a case of making FreeMiNT mount read-only when it initializes. If FreeMiNT is internally opening/writing files before the filesystem it wants to write to is mounted read-write, then that's another FreeMiNT problem that should be fixed.

Modern Unix's don't hide it, it's still all in the scripts, just as ours would be, and that's entirely dependent on the users choice of boot up scripts, whether that be SpareMiNT, Gentoo, and anything else. It's the sysadmin job to ensure the scripts are running correctly.

atic-atac:
@mikrosk you can't revert the patch, not only does it fix the > 16 partition limit it also lets the underlying hard disk driver know of other limitations that have changed.

Why is re-mounting the FAT or TOS partitions read-only problematic when FreeMiNT starts ?

There are MS_READONLY and MS_REMOUNT flags already, but we don't seem to be using MS_REMOUNT at this time.

atic-atac:
@mikrosk you can't revert the patch, not only does it fix the > 16 partition limit it also lets the underlying hard disk driver know of other limitations that have changed.

Why is re-mounting the FAT or TOS partitions read-only problematic when FreeMiNT starts ?

There are MS_READONLY and MS_REMOUNT flags already, but we don't seem to be using MS_REMOUNT at this time.

The initramfs idea is another solution, but that probably involves as much work as making fsck tools run before MiNT.

atic-atac:
@th-otto @mikrosk There's also these controls already in FreeMiNT to control filesystem readonly/readwrite modes...

# define V_CNTR_WP      (('V'<< 8) | 100)       /* MiNT, write protect control */
# define V_CNTR_WB      (('V'<< 8) | 101)       /* MiNT, write back control */

fsetter looks like it uses them. But you'll notice these controls in each filesystem driver to handle mounting read-only.

atic-atac:
@th-otto @mikrosk There's also these controls already in FreeMiNT to control filesystem readonly/readwrite modes...

# define V_CNTR_WP      (('V'<< 8) | 100)       /* MiNT, write protect control */
# define V_CNTR_WB      (('V'<< 8) | 101)       /* MiNT, write back control */

fsetter looks like it uses them. But you'll notice these controls in each filesystem driver to handle mounting read-only.

skaftetryne:
@atic-atac 

> FreeMiNT should remount them read-only when it starts, execute the startup scripts as any standard unix system would do, and run fsck during those scripts, and which point the scripts then mount read/write and continue.

But again, MiNT IS NOT UNIX. I got this feeling that some people enjoy watching lots of text flying by during boot, as this makes it looks like a serious system... Don't forget that a large portion of the users doesn't use these "standard" scripts at all, as they - for lots of users - does nothing but slow down boot, eat memory and complicate configuration changes.

Exactly *what* problem does such an approach solve? That it's not "unix"? Do we really need all this just to get the filesystems checked? Is there no other way than the unix-way?

MiNT is not unix. We can do things much simpler and more userfriendly than simply copying 40 year old concepts.

1. Filesystems can be checked before MiNT loads the fs drivers. But that still leaves the "problem" with FAT-drives.
2. Filesystems can be checked before MiNT is started. But same "problem" as in 1. On the plus side - this benefits users of TOS, Geneva and MagiC too.
3. Dirty bit doesn't have to be set until the filesystem is actually written to. That's the entire point of the dirty bit. The fs is not dirty just because it's "mounted".

Personally I like 3 best. It's completely transparent to the user.

skaftetryne:
@atic-atac 

> Why is re-mounting the FAT or TOS partitions read-only problematic when FreeMiNT starts ?

First of all - they will most likely already have been written to by TOS before MiNT is started. So they are already "dirty". And MiNT writes mint.ini and log entries before any script or program can be started from mint.cnf.

atic-atac:
@skaftetryne That's the point of FreeMiNT re-mounting read-only, it would flush the disk caches, before mounting read-only. Nothing would be dirty when FreeMiNT re-mounts read-only.

atic-atac:
@skaftetryne Also, remember there is a situation where the dirty bit is set before MiNT boots, when the system is shutdown uncleanly.

mikrosk:
@atic-atac so you're saying that the typical init script should execute a binary which would call `Dcntl (V_CNTR_WP, "DRIVE:", 0);` ?

I'm confused by one thing you've written:

> all partitions that your hard disk driver knows about **will be mounted read-write** at boot. FreeMiNT should remount them read-only when it starts

mounted by whom? Why FreeMiNT can't just mount the drives as read-only as soon as the file system driver is loaded? Or is there a way in TOS to "mount" something so it has to be unmouted/remounted first?

@skaftetryne

> That's the entire point of the dirty bit. The fs is not dirty just because it's "mounted".

Actually, there's some information written, at least in case of ext2 -- like mount time or number of mounts as soon as it is mounted.

For me personally, the only obstacle I see with having this "remount as rw later" approach is that we would need make sure that the "enable rw" binary (see beginning of this post) is executed from mint.cnf by default in the distribution archive and really, really, really make sure that everybody understand it must be executed from there (perhaps even writing a warning / alert during the boot if mint.cnf parsing leaves with all partitions r/o).

mint.ini and logs can be solved quite easily by writing it to memory (ramfs, remember?) first and then "flush" if filesystems are proven to be safe to write to.

atic-atac:
@mikrosk That's what I said, the FreeMiNT kernel would mount the filesystems by default as MS_RDONLY when the filesystem drivers are loaded. But an external binary is executed later to re-mount them read/write, clearing the MS_READONLY bit through the V_CNTR_WP/WB calls. aka the "mount" command on a unix system.

We can add an option to FreeMiNT startup options (using shift) to make FreeMiNT mount read/write if necessary too.

mikrosk:
> We can add an option to FreeMiNT startup options (using shift) to make FreeMiNT mount read/write if necessary too.

Yes, this is a neat idea. It could have different levels -- remount as 'rw' as soon as possible (for instance before mint.cnf parsing), before `GEM=`/`INIT=` or not at all.

skaftetryne:
@atic-atac 

> That's the point of FreeMiNT re-mounting read-only, it would flush the disk caches, before mounting read-only. Nothing would be dirty when FreeMiNT re-mounts read-only.

In that case, why can't fsck just flush the disk cache itself? Introduce a new systemcall (or use Fchmod) to let fsck set the entire disk to read-only and then let if flush the cache. Then there is no need to mess around with mounting.

skaftetryne:
@mikrosk 
>For me personally, the only obstacle I see with having this "remount as rw later" approach is that we would need make sure that the "enable rw" binary (see beginning of this post) is executed from mint.cnf by default in the distribution archive and really, really, really make sure that everybody understand it must be executed from there (perhaps even writing a warning / alert during the boot if mint.cnf parsing leaves with all partitions r/o).

I can assure you that users will get confused by this. Just look at some of the questions on the mailing-list or atari-forum - a lot of stuff that we find obvious is voodoo to many users.

Why can't fsck take care of this itself? I.e. set filesystem to read-only, flush the cache, do it's work, set filesystem to r/w? Why leave this to separate programs?

th-otto:
@skaftetryne i think you missed an important point. The reason why fsck checks the file system state flag is to determine wether it has to do something. If the filesystem it clean, it usually exits immediately (unless you have configured to check the fs anyway after some amount of time and/or mounts).
Due to the current behaviour, fsck already sees the filesystem as dirty. It cannot distinguish wether that was because of the bug, or due to some improper shutdown. Mounting the fs readonly at this time is too late. Beside that, there is no explicit mount call in MiNT.

atic-atac:
@skaftetryne There's a 50/50 choice here. People who are used to Unix and those who are used to TOS. I'd hazard a guess, anyone using MiNT by now is already used to some level of Unix, and if they are not, I suspect they'd have been turned off from MiNT a long time ago.

We already have a "flush the cache tool" in the command "sync". There would be a command for mounting filesystems called "mount", why integrate these tools into "*fsck" ??? You'd certainly confuse those from unix backgrounds as we use most of the open source tools within MiNT to manage ext2fs.

But also, it doesn't stop someone from adding what you are asking to each *fsck tool. Then people have choices. The option in FreeMiNT for readonly/readwrite is also a choice.

skaftetryne:
@atic-atac 
> There's a 50/50 choice here. People who are used to Unix and those who are used to TOS. I'd hazard a guess, anyone using MiNT by now is already used to some level of Unix, and if they are not, I suspect they'd have been turned off from MiNT a long time ago.

Yes, that's your guess. But if *I* should guess by the number of downloads of VanillaMiNT (which is also shipped with every Firebee) there are quite a lot of people that use MiNT without any sort of "unix" at all.

skaftetryne:
@th-otto 
> i think you missed an important point. The reason why fsck checks the file system state flag is to determine wether it has to do something. If the filesystem it clean, it usually exits immediately (unless you have configured to check the fs anyway after some amount of time and/or mounts).

No, I'm fully aware of the problem. It's just that I think it's a bad choice to "fix" this by making MiNT harder to use.

e2fsck worked for 15 years without this issue.

atic-atac:
@skaftetryne If you want to ship a modified e2fsck program with VanillaMiNT, that's absolutely fine. Any changes that happen to FreeMiNT would be optional so shouldn't affect the operation of VanillaMiNT. You'd have the choice to still mount read/write and do whatever you need.

Personally, I don't think it's the right solution. The dirty bit in the FS could already be set, so you don't want to give FreeMiNT a chance to start writing to a dirty filesystem. It should be mounted read-only.

Also, keeping modifying e2fsck doesn't appeal to me, so I'd be happy with a new "mount" program that re-mounted the filesystem read/write after the fsck tool have been executed. This new mount program would be under tools/ and be alongside our mount_nfs tool already.

To claim it worked for 15 years, doesn't change the fact I think FreeMiNT has been broken during that time by continually mounting the FS read/write when it can be dirty.

mikrosk:
@skaftetryne believe me, I'm one of the TOS/GEM friends here (never really liked the Unix side of FreeMiNT except for easy porting when compiling stuff with gcc and sure as hell I didn't like all that fsck crap for "no" obvious reason for a beginner) but I still fail to see what's the problem here, esp. after @atic-atac's suggestion to offer auto mount as rw (say, after mint.cnf parsing). How that would upset anyone?

mikrosk:
@atic-atac just a little heads up. I've looked into `V_CNTR_WP` processing and ro mounting. It could easily work except the fact it wouldn't solve this issue at all. ;) The problem is that every `*_dskchng()` function purposely ignores the read only flag. Every other write function checks for it and return but `dskchng`? No way. Flushed to disk no matter what.

So I start to be a little suspicious about this whole `dskchng()` handling, starting with @vinriviere's smart question as why we're flushing anything at all and ending with this very strange "write no matter what".

Because you know... the easiest fix, which would make also @skaftetryne happy ;), would be just... removing that flush code.

mikrosk:
Btw, while reading the kernel documentation, by dskchng() it mentions explicitly only invalidating, exactly as @vinriviere pointed out.

All docs I could find mention `Mediach()` reporting as something related to removable medias. Can anyone imagine a scenario where a "media" change would occur and we would want to write data back to disk and *then* invaliding the cache? I can't believe everybody in (Free)MiNT's history would overlook such a blunt error.

atic-atac:
@mikrosk can you give a pointer to the flushing cache around dskchng ? 

I've taken a quick look and see some invalidation code. But nothing pops out that flushes the cache.

mikrosk:
@atic-atac it's the `bio.sync_drv (s->di)` call. This leads to `bio_wb_queue()` and this writes everything back to disk via `Rwabs()` (either the original or XHDI).

atic-atac:
@mikrosk ok I see now, and fatfs.c does it too. So yes, I think this is wrong. We need to review each of the filesystem driver code to see what each does.

mikrosk:
@atic-atac minixfs does it too, others are fine (empty `dskchng` or not handling any cache).

After spending nearly a week on this code I think someone for some reason abused `dskchng` as a kind of flush function. You can perfectly see that in `unmount` -- it sets the valid state flag and calls `dskchng` as it was the most natural thing to do. It has been like that for 17 years.

atic-atac:
@mikrosk ok, it sounds like unmount needs updating to flush the cache instead.

atic-atac:
@mikrosk and thanks for digging into this!

atic-atac:
@mikrosk I just checked minixfs and it does it right, there is no cache flush in m_dskchng, just in m_unmount, then it calls m_dskchng after calling bio_sync.

atic-atac:
I can only see fatfs.c & ext2sys.c being a problem. Below patch.....

diff --git a/sys/fatfs.c b/sys/fatfs.c
index 0813c65..f7da67b 100644
--- a/sys/fatfs.c
+++ b/sys/fatfs.c
@@ -7691,9 +7691,6 @@ fatfs_dskchng (int drv, int mode)
 
 	FAT_DEBUG (("fatfs_dskchng: invalidate drv (change = %li)", change));
 
-	/* I hope this isn't a failure */
-	bio.sync_drv (DI (drv));
-
 	/* invalid all cookies */
 	{
 		register long i;
@@ -7768,6 +7765,9 @@ fatfs_sync (void)
 static long _cdecl
 fatfs_unmount (int drv)
 {
+	/* I hope this isn't a failure */
+	bio.sync_drv (DI (drv));
+
 	if (CLEAN (drv))
 		clean_flag (drv, CLEANFLAG_SET);
 
diff --git a/sys/xfs/ext2fs/ext2sys.c b/sys/xfs/ext2fs/ext2sys.c
index 02e951b..332b2f3 100644
--- a/sys/xfs/ext2fs/ext2sys.c
+++ b/sys/xfs/ext2fs/ext2sys.c
@@ -2150,12 +2150,6 @@ e_dskchng (int drv, int mode)
 
 	DEBUG (("Ext2-FS [%c]: e_dskchng (mode = %i): invalidate drv (change = %li, memory = %li)", drv+'A', mode, change, memory));
 
-	/* sync the inode cache */
-	sync_cookies ();
-
-	/* sync the buffer cache */
-	bio.sync_drv (s->di);
-
 	/* free the DI (invalidate also the cache units) */
 	bio.free_di (s->di);
 
@@ -2296,6 +2290,12 @@ e_unmount (int drv)
 		DEBUG (("can't unmount cleanly"));
 	}
 
+	/* sync the inode cache */
+	sync_cookies ();
+
+	/* sync the buffer cache */
+	bio.sync_drv (s->di);
+
 	e_dskchng (drv, 1);
 
 	return E_OK;

mikrosk:
@atic-atac you're right, I was too quick to write minixfs off, it's done (as the only one) properly. However, be careful, fatfs has to clean the flag first, then sync()

I've looked into usage of the force flag: `dskchng()` is used only here: `sys_d_lock()` (harmless as all three fs drivers have `FS_EXT_1` set) and here `shutdown()` (ditto) and they are properly described as invalidate functions. So it should be OK.

atic-atac:
@mikrosk gotcha, i've flipped the clean flag first for fatfs.c.

Preferably, I'd rather duplicate the code for invalidation into unmount, as I don't like calling other filesystem functions from another as it duplicates messages and can be confusing.

mikrosk:
@atic-atac agreed, I meant that your patch should be OK because no other kernel function (apart from xfs `unmount`) depends on the flushing functionality in `dskchng`. I'll test it tomorrow and let you know.

atic-atac:
@mikrosk ok, full patch...

```
diff --git a/sys/fatfs.c b/sys/fatfs.c
index 0813c65..3ea2818 100644
--- a/sys/fatfs.c
+++ b/sys/fatfs.c
@@ -7691,9 +7691,6 @@ fatfs_dskchng (int drv, int mode)
 
 	FAT_DEBUG (("fatfs_dskchng: invalidate drv (change = %li)", change));
 
-	/* I hope this isn't a failure */
-	bio.sync_drv (DI (drv));
-
 	/* invalid all cookies */
 	{
 		register long i;
@@ -7771,7 +7768,33 @@ fatfs_unmount (int drv)
 	if (CLEAN (drv))
 		clean_flag (drv, CLEANFLAG_SET);
 
-	fatfs_dskchng (drv, 1);
+	/* I hope this isn't a failure */
+	bio.sync_drv (DI (drv));
+
+	/* invalid all cookies */
+	{
+		register long i;
+		for (i = 0; i < COOKIE_CACHE; i++)
+		{
+			register COOKIE *c = &(cookies[i]);
+			if (c->dev == drv)
+			{
+				if (c->name)
+				{
+					c_del_cookie (c);
+				}
+			}
+		}
+	}
+
+	/* free the DI (also invalidate cache units) */
+	bio.free_di (DI (drv));
+
+	/* invalidate the BPB */
+	BPBVALID (drv) = INVALID;
+
+	/* free the dynamically allocated memory */
+	kfree (BPB (drv)); BPB (drv) = NULL;
 
 	return E_OK;
 }
diff --git a/sys/xfs/ext2fs/ext2sys.c b/sys/xfs/ext2fs/ext2sys.c
index 02e951b..6e8c76d 100644
--- a/sys/xfs/ext2fs/ext2sys.c
+++ b/sys/xfs/ext2fs/ext2sys.c
@@ -2150,12 +2150,6 @@ e_dskchng (int drv, int mode)
 
 	DEBUG (("Ext2-FS [%c]: e_dskchng (mode = %i): invalidate drv (change = %li, memory = %li)", drv+'A', mode, change, memory));
 
-	/* sync the inode cache */
-	sync_cookies ();
-
-	/* sync the buffer cache */
-	bio.sync_drv (s->di);
-
 	/* free the DI (invalidate also the cache units) */
 	bio.free_di (s->di);
 
@@ -2296,7 +2290,26 @@ e_unmount (int drv)
 		DEBUG (("can't unmount cleanly"));
 	}
 
-	e_dskchng (drv, 1);
+	/* sync the inode cache */
+	sync_cookies ();
+
+	/* sync the buffer cache */
+	bio.sync_drv (s->di);
+
+	/* free the DI (invalidate also the cache units) */
+	bio.free_di (s->di);
+
+	/* clear directory cache */
+	d_inv_dir (drv);
+
+	/* clear inode cache */
+	inv_ctable (drv);
+
+	/* free allocated memory */
+	kfree (s->sbi.s_group_desc, s->sbi.s_group_desc_size);
+	kfree (s, sizeof (*s));
+
+	super [drv] = NULL;
 
 	return E_OK;
 }

```

mikrosk:
It works. I had to lower BIO_FORCE -> BIO_DEBUG so user wouldn't be confronted with this scary message.

There's one issue, though. `dosfsck` doesn't reset the dirty flag. So if you don't cleanly shutdown the OS, you'll never get rid of the "unchecked file system" message on FAT. Strange thing is that the mint patch for dosfsck 2.10 contains detection of it but not (re)setting. [Source code](https://github.com/freemint/sparemint/blob/master/sparemint/SRPMS/dosfstools-2.10-1.src.rpm) is just mess, I don't dare adding it there. Upstream had included support for the dirty flag long after 2.10 and it's not totally trivial to create an up to date mint patch there.

Rustynutt:
Have seen in the forums mikrosk plead more involvement with MiNT development. 
Have to admit, this process much more structured. In the past have frequented Sourceforge, not because it a better platform, I'm just old, with Github being relatively new.
The last MiNT setup I'd solidly utilized is with an Afterburner, maybe V1.14.
The past couple weeks have dived into recovering a SpearMiNT installation configured on a Falcon (down to 4) I'd managed to disrupt. 
Vanilla MiNT seemed the quickest way to regain access to the extends and FAT32 partitions to archive the extensive setup. I see RPM, X installed.
The Vanilla MiNT install, however simple, was never stable enough for comfort. Followed wongck's MiNT installation guide, very extensive, up to using tools to format the ext2 and FAT32 fs. Granted, he said it's an older guide. Since Vanilla MiNT was up, connected a secondary IDE slave, used HD Driver V11.X to setup a 1GB TOS partition, 34GB ext2 and 45GB FAT32 partitions.
Then, what many users that have a case of the ass installing MiNT, came to a questionable impasse. While not a programmer, hardware expert, am methodical.
Where wongck provides a link to snag the formatting packages, the naming convention changed 3 times in as many short paragraphs. Kind of had to guess which was the correct tool for each task. Even went as far as trying to find the softs in existing installations, not present. That lead me on a chase, landing on some Unix page where the maintainer has V4 of the tool. I'd missed the part I wasn't in Oz anymore (TOS world), even if a compiler is installed, would do better courting  Amanda Redman. Beaten back to the machine in front of me yet wondering if I'm utilizing an outdated program, used the formatting tools provided via wongck's link. ext2, ttp renamed, made quick work of the partition.  He also states it's better to use the tool under MiNT to format the FAT32 partitions as HD Driver has a bug. Well, it failed to begin the format process. Tried a couple times, lastly, NOT using the Unix package fatfs tool.
This process could go one 3x as much text, basically began having problems with the ext2 partition, mitigating to the FAT32 as well.
Firstly, this "commit" has taught me to look in the SpareMiNT package for the disk tools. I did t see them I. The MiNT packages mikrosk has out up here don't seem to contain the tools either, leading me to think they are not part of the MiNT archive, for one reason or another.
Second thing I learned, more like came to the realization of, there possibly isn't one MiNT kernal identical to another, each user with resources enough can make changes, much as discussed in this commit. This I discovered, taking the MiNT binary from mikrosk's archive, and placing it in the Vanilla MiNT installation. Not only a file size difference, there's more going on in the Github version.
Breathe :)
Wife had a knee replacement yesterday, be a couple days before setting in front of the Falcon.
For SCSI, have a SCSI2SDV6F, Aztec CF, Accent CF, SCSI CD, SCSI HDD. On IDE can set up 2.5" laptop HDD, 3.5" HDD, IDE CF (as well as dual CF), IDE to 50 pin slimline DVD, DOM IDE, IDE 2 SATA HDD, NetUSB (but so darn slow) USB DVD, USB stick, original 80mHz CT60 w/128mb RAM, SuperVidel (V11 scores), SVLana (online), EtherNat (untested), Eiffel (non-display) and Didier's original CTTOS 1.03c.
Guess what I'm getting at is i can happily test a number of configurations, being "Hello World", in basic my limit other than 5 axis CNC programing, trig and geometry strong points.
It's either this, or turn to the dark side, but am a believe MagiC should be left to the MAC's :)
My next step is to jump into the SpearMiNT package, get the 3.5" HDD configured and back up the existing package and go from there.
I want to compile some stuff! :)

PR #47, opened 2017-06-07
Fix debug usage in aranym.xfs (nfclipbrd driver)

th-otto:


mikrosk:
If you're cleaning source files, please separate the change (in this case moving stuff from main.c to a separate file) -- I have really hard time to understand what actually this PR fixes.

th-otto:
I can split the PR into two commits if that makes it clearer, but they should go together. The splitting was done to eventually allow the driver to be compiled into the executable, just like the hostfs.

mikrosk:
Yup, two commits are fine.

Issue #45, opened 2017-06-07
Daynaport driver (scsilink.xif) fails to compile


th-otto:
When do a fresh clone of the repo, then
$ export M68K_ATARI_MINT_CROSS=yes
$ make

i get

make[4]: Entering directory `/home/sebilla/freemint/sys/sockets/xif/daynaport'
m68k-atari-mint-gcc -I../.. -I../../.. -D__KERNEL_XDD__ -Werror -mshort -g -O2 -fomit-frame-pointer -Wall -Wmissing-prototypes -Wshadow -Wpointer-arith -Waggregate-return -c scsilink.c -o scsilink.o
scsilink.c: In function 'driver_init':
scsilink.c:192:2: error: passing argument 3 of 's_system' makes integer from pointer without a cast [-Werror]
../../../libkern/kernel_xfs_xdd.h:671:13: note: expected 'ulong' but argument is of type 'void *'
cc1: all warnings being treated as errors

Strangely, the last automatic builds via travis do not seem to be affected by this. Also notice that there is no -m68xx switch in the compile command, although the Makefile specifies CPU = 020-60. As already mentioned on the ML, any CPU = statement in the Makefile(s) must be done *before* RULES is included.

mikrosk:
I guess you're using gcc > 4.6.4.

However I can reproduce the lack of `-m68020-60`, any idea what has caused it?

mikrosk:
> it's bad positionioning of the CPU= statement

Hmm, there must be something more to it. Look at `xdd/dsp56k`, it's nearly the same Makefile and it works properly.

th-otto:
no, just normal one:

m68k-atari-mint-gcc --version
m68k-atari-mint-gcc (GCC) 4.6.4 (MiNT 20130415)

mikrosk:
I'm using the same and can compile it fine.

th-otto:
Hm, that must be the NULL argument. Maybe some header slips in that defines it as ((void *)0), which it should be rather than just 0. In any case, using NULL when the function expects ulong is simply wrong. OTOH, that arg2 of s_system should actually be a pointer to a long.

th-otto:
> Hmm, there must be something more to it. Look at xdd/dsp56k, it's nearly the same Makefile and it works properly.

not for me:

$ cd sys/xdd/dsp56k
$ rm dsp56k.o
$ make dsp56k.o
[...]

PR #48, opened 2017-06-08
Move CPU overrides in Makefiles to the top

th-otto:
The CPU variable is used in RULES to set the MODEL variable,
and must therefore be set before that file is included

This should solve the issues discussed in #45


mikrosk:
Cool, thanks for looking into this! How come files like `sys/usb/src.km/ucd/netusbee/Makefile` do not need to be changed? I also see RULES included before the CPU definition.

mikrosk:
Thanks for looking into this! It looks good (I overlooked some changes before).

sys/arch/magic/Makefile
```diff
@@ -47,8 +47,8 @@ $(TARGET): $(OBJS) $(LIBKERNSTMP)
 ifeq ($(CROSS),yes)
 
 # for cross-compiling use this: (assumes you have perl and cnm)
-magic.i: $(TARGET)
-	perl genmagic.pl > $@
+magic.i: genmagic.o
+	perl genmagic.pl > $@.$$$$ && mv $@.$$$$ $@ || { $(RM) $@.$$$$; false; }
```
mikrosk: What does this wizardry means? :)

th-otto: That wizardry writes to a tmp file and moves it only when it finished successfully. Otherwise when using parallel makes, the redirection wil create it in one make, and the 2nd make will think it is up to date and continue with an empty output file because the first command isn't finished yet.

th-otto: BTW any chance of getting that (and the other PRs)  merged in the near future?

mikrosk:
@th-otto: patience. :) @atic-atac is quite busy these days but eventually each PR gets reviewed.

PR #49, opened 2017-06-09
Support file locking in hostfs

th-otto:
File locking may require to have to suspend the process,
so the hostfs part in the emulator will never be able to
emulate file locking, since it does not have any notice of
MiNT processes. This must be done by hostfs.xfs.
File locking is eg. used by rpm, which will fail without this.


atic-atac:
There's more to this than file locking. There's date code being modified. Can you elaborate ?

th-otto:
Oops. Should have been separate commit. Is fixed now. The first commit still has some diffs against dosfile.c that should actually have been part of #50, sorry for that.

PR #50, opened 2017-06-09
Add gnu attributes to ksprintf/kvsprintf and fix format strings

th-otto:


mikrosk:
Your change emits a warning which is treated as error.

th-otto:
Not any more ;)

PR #64, opened 2017-08-14
Fix the SR saving for the EtherNAT handler

### LINKED ISSUE #63: Questionable CPU interrupt mask handling in USB stack
While reviewing the code of FreeMiNT's USB stack I found some code that I believe is wrong. More precisely, the handling of the 68k's interrupt mask in the EtherNat driver. The pack_fifo() and unpack_fifo() functions increase the interrupt mask (to 6), in order not be interrupted during their work. That's OK, of course. 

However, they call set_old_int_lvl() (see https://github.com/freemint/freemint/blob/1d45d4f8216b5f780577d5f15fbe0e3651273438/sys/usb/src.km/ucd/ethernat/isp116x-hcd.c#L636 for an example) when they are finished. Other than one might suspect given the function name, set_old_int_lvl() does _not_ restore the previous interrupt mask. Instead it takes a word from the top of the stack and uses that as new mask, as can be seen here: https://github.com/freemint/freemint/blob/1d45d4f8216b5f780577d5f15fbe0e3651273438/sys/usb/src.km/ucd/ethernat/ethernat_int.S#L66. 

This only makes sense when used at the very beginning of an interrupt handler, where in fact the previous interrupt mask (or rather: the previous value of SR) is on the stack. In the EtherNat USB driver however, it will just set the interrupt mask to an arbitrary value, depending on what happens to be on the stack. I don't know if something in the MiNT kernel will reset the interrupt mask to the correct value, however, I think the current code is wrong.

mikrosk:
This fixes #63. Contributed by Christian Zietz.

### DIFF
--- a/sys/sockets/xif/ethernat/ethernat_200Hzint.S
+++ b/sys/sockets/xif/ethernat/ethernat_200Hzint.S
@@ -74,7 +74,6 @@ oldSR:	ds.w	1
 
 | Sets interrupt level to what was in the SR
 _set_old_int_lvl:
-	move.w	(sp),oldSR
 	andi.w	#0x0f00,oldSR		|just keep the int lvl
 	move.l	d0,-(sp)
 	move.w	sr,d0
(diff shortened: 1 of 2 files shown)


mikrosk:
I've noticed that the network driver (ethernat.xif) shows the same bug. Surprisingly, svethlana.xif is fine because it doesn't call `set_old_int_lvl()` at all.

PR #83, opened 2018-04-01
Fix potential address error crash in USB storage driver

### LINKED ISSUE #81: Potential address error crash in USB storage driver
Hello,

in _TOSONLY_ mode, the USB storage driver uses a potentially dangerous way to get the XHDI cookie: 
https://github.com/freemint/freemint/blob/1196c6f665b6eed314b419ce560f445ebaeae5a1/sys/usb/src.km/udd/storage/xhdi.c#L1036

The problem lies in the cast: `*((long *)"XHDI")` gets compiled into

```
xhdi_string:
dc.b "XHDI",0
[...]
move.l  xhdi_string,d1
```

..., i.e. the code actually reads the memory location `xhdi_string` instead of loading the immediate value #0x58484449 (=XHDI) into D1. However, the compiler does not care about the alignment of `xhdi_string` in memory. It is a char array, after all. But the 68000 can only read longwords from even addresses. (This was changed iirc in the 68020.) 

So, if by chance `xhdi_string` is at an odd memory location, the driver crashes with 3 bombs = address error. It is pure luck that the driver works. By introducing some unrelated changes I changed the location of `xhdi_string`, exposing this crash.

The code should be modified (in several places!) not to use the dangerous cast.

mikrosk:
Resolves #81

### DIFF
--- a/sys/usb/src.km/global.h
+++ b/sys/usb/src.km/global.h
@@ -23,6 +23,7 @@
 #ifndef _global_h
 #define _global_h
 
+#include "cookie.h"
 #include "mint/module.h"
 #ifndef TOSONLY
 #include "libkern/libkern.h"
@@ -156,12 +157,6 @@ static inline long strlen (const char *scan)
 
 /* cookie jar definition
  */
-
-struct cookie
-{
-	long tag;
-	long value;
-};
 
 #define _USB 0x5f555342L
 #define CJAR ((struct cookie **) 0x5a0)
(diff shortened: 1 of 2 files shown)


th-otto:
You should use include <sys/cookie.h> here. <mint/cookie.h> is from the mintlib, but the kernel sources should not depend on it.

mikrosk:
True. I was fooled by xaloader which is using `<mint/cookie.h>` but that is of course correct.

PR #97, opened 2018-08-22
[RFC] Add multi target compile for /tools

DavidGZ:
This pull request is "request for comments" before I push those commits into the repository.

I still need to change Travis scripts before submitting these changes.

I've only generated three binaries for every tool (000, 020-60, ColdFire), as we do in the snapshots. I know there are people who likes to have a binary for every single CPU supported, I like something more simple, in my opinion those three target are enough , but this can be discussed here.


mikrosk:
I guess @th-otto is the right person to comment on the technical side. I welcome this change and don't worry about Travis scripts, I'll fix them after this has been committed. This will clean up a lot of hacky stuff which sometimes even I have trouble understand. :)

The only thing which saddens me is this inconsistency - I'm all in for reducing the number of CPU targets but now some parts will have all of them and some just three. I still have a patch somewhere which reduces the number of targets to what you have done here but I'm not sure whether that time has come (that would also mean the end of the 'old school' builds).

Also, some of the tools are going away sooner or later (HypView comes into my mind), so only the pure mint-related tools will be kept.

th-otto:
I haven't looked trough all the details yet, but assuming that the changes all follow the same scheme, generally looks good to me. Of course the build failures should be fixed ;)
Only small thing: in some directories (eg. mktbl) you include a MKTBLDEFS file, which defines $(TARGET) according to the toplevel target. If i'm not wrong, that file is not needed in the toplevel and could just be integrated into Makefile.objs.

DavidGZ:
>Only small thing: in some directories (eg. mktbl) you include a MKTBLDEFS file, which defines $(TARGET) according to the toplevel target. If i'm not wrong, that file is not needed in the toplevel and could just be integrated into Makefile.objs.

Sorry Thorsten, I'm not sure if I understand what you mean. Which file could be integrated in Makefile.objs? MKTBLDEFS? If so, the file Makefile in the top level needs MKTBLDEFS to be included  too for the variable mktbltargets.

th-otto:
> Which file could be integrated in Makefile.objs? MKTBLDEFS?

Yes. Same for the other directories of course, that was just an example.

>the file Makefile in the top level needs MKTBLDEFS to be included too for the variable mktbltargets.

It's  a bit difficult to see without applying the patch, but i don't think so. MKTBLDEFS defines $(TARGET) and does not need mktbltargets, but the toplevel Makefile does not reference that variable anymore after the changes. It uses $(mktbl) and $(mktbltarget) instead. The line with mktbltargets should indeed then be moved to the toplevel Makefile instead, it is not used by Makefile.objs

But might as well do that change in a 2nd step, when it is easier to see, and you can actually try whether that works.

DavidGZ:
>It's a bit difficult to see without applying the patch, but i don't think so. MKTBLDEFS defines $(TARGET) and does not need mktbltargets, but the toplevel Makefile does not reference that variable anymore after the changes. It uses $(mktbl) and $(mktbltarget) instead. The line with mktbltargets should indeed then be moved to the toplevel Makefile instead, it is not used by Makefile.objs
But might as well do that change in a 2nd step, when it is easier to see, and you can actually try whether that works.

OK, I understand what you mean now. In any case those changes are beyond these patch series because that should be change through all the kernel source tree for consistency.

DavidGZ:
>> The only thing which saddens me is this inconsistency - I'm all in for reducing the number of CPU targets but now some parts will have all of them and some just three. I still have a patch somewhere which reduces the number of targets to what you have done here but I'm not sure whether that time has come (that would also mean the end of the 'old school' builds).

>I'm going to open a dialog in the mailing list and see what comes from it.

I've added the rest of the CPUs to the pull request, if the time arrives when we decide to keep only 3 targets we'll remove them again.

@mikrosk, as soon as it's a good time for you to do the changes in the travis scripts I'll rebase and merge this pull request.

mikrosk:
All right, as soon as I finish my gcc adventure I'll do this one.

th-otto:
Your PR currently fails to build (with OLD_BUILD at least), because Travis now limits the log file to 10000 lines. I had similar problems already in aranym with the macos build, but ther i could easily solve it by silencing down xcodebuild a bit. In our case, i have currently no idea to prevent that other than redirecting the whole output to a file, but then you won't see anything in the travis log. Any ideas?

th-otto:
Opened a new issue for the above problem since its unrelated to the PR.

DavidGZ:
> Your PR currently fails to build (with OLD_BUILD at least), because Travis now limits the log file to 10000 lines. I had similar problems already in aranym with the macos build, but ther i could easily solve it by silencing down xcodebuild a bit. In our case, i have currently no idea to prevent that other than redirecting the whole output to a file, but then you won't see anything in the travis log. Any ideas?

It happened after I solved the conflicts that appeared after your latest commits for COPS, I had no idea why it was failing, I was waiting Miro to come to the rescue :-)

BTW after I solved those conflicts a new [commit](https://github.com/freemint/freemint/pull/97/commits/44293b0bf1be804019d2c97517f9d9943af564d6l) was added to the pull request, I try a couple of times but I don't know how to avoid this, I guess when doing a rebase & merge git will remove that commit, but I won't be sure until I do it and I wouldn't like to "pollute" the repository's git history. Can somebody confirm this?

th-otto:
I think it was already failing before, we just didn't notice until Miro fixed the script to check the exit status. Your PR just happens to trigger this limit, because the tools are now compiled multiple times, thus producing more output.

>BTW after I solved those conflicts a new commit was added

Just a merge commit, yes, that's normal.

mikrosk:
@DavidGZ just a minor complain, the 020-60 target directory should be named 'compile_02060' according to your previous naming schemes.

mikrosk:
Two more things:
- please put `copstargets =` on one line. yes, it looks ugly but makes my `sed`ing much easier :)
- please extend your PR also to `sys/usb/src.km/loader` and `xaaes/src.km/xaloader`

With that done I can simplify a lot of things in the build scripts and make it right.

DavidGZ:
@mikrosk OK, I'll take a look at those issues but after next Monday, I'm busy till then.

mikrosk:
No problem David, there's no rush.

DavidGZ:
@mikrosk your requests from above are now complete, as soon as you can we can proceed with the merge.

mikrosk:
Yup, looks good. Will take a look tomorrow. Thanks!

mikrosk:
@DavidGZ @th-otto guys as I'm testing my scripts, I have realised we are pretty close to fixing #12. However there are two issues:
- if invoking "make CPU=030" in, say, xaloader, xaloader will be built in all .compile_* directories with `-m68030` flag
- if invoking "make 030" in a toplevel directory (and also "sub toplevel" like tools, xaaes etc) it doesn't work

While perhaps the latter case is more related to #12, the former could be maybe addressed here, while we are changing tons of files anyway? As `CPU` is used as the cpu id in all the *DEFS files (and is overridden if `CPU` is specified during `make` invocation), I can imagine only one way to fix this and this is to introduce something like "make TARGET_CPU=xxx" and this `TARGET_CPU` would be then compared with the actual `CPU` value (so if `TARGET_CPU` is set - build only if `TARGET_CPU` == `CPU` else build `CPU`). 

Other approach could be that we handle `CPU = 000` in *DEFS in a different way, something like `CPU ?= 000` and then somehow compare & skip building this target if CPU isn't really 000.

What do you think?

mikrosk:
As a proof of concept I tried this (whlmoose):

MOOSEDEFS:
```Makefile
ifeq ($(moose),040)
TARGET = moose.adi
CPU ?= 040
ifneq ($(CPU),040)
SKIP_THIS_BUILD = yes
endif
MOOSEDEFS +=
endif
```
Makefile:
```Makefile
ifeq ($(SKIP_THIS_BUILD),yes)
realtarget =
endif
```

And it worked. However it's too verbose, so if you can come up with a more elegant solution I'm all ears. :)

th-otto:
> if invoking "make CPU=030" in, say, xaloader, xaloader will be built in all .compile_* directories with -m68030 flag

That's why you should not do that ;)

> can imagine only one way to fix this and this is to introduce something like "make TARGET_CPU=xxx" 

What for? That would make things even more complicated than they already are.

CPU= is passed in the Makefiles when building specific targets, ie when building the aranym version it is set to 040. Allowing that to be specified also on the commandline, will only break things and does not make sense. You don't want a milan version that is compiled for plain m68k (or maybe even coldfire). It is already explained what targets are available, we should stick to that.

th-otto:
You could fix #101 also by building the matrix based on target instead of CPU type. Or by building the kernel and the tools separately. Or by moving apps like cops, toswin, hypview etc. into separate repos, imho they don't belong into the kernel repo anyway.

> with the approach presented above the Milan target would be skipped if make was invoked as make CPU=000.

But only by introducing redundant logic about which targets are build for which CPU type.

th-otto:
BTW currently only the OLD_BUILD=1 is reported as failing. But the log for OLD_BUILD=0 also ends after 10000 lines. Could it be that there is also a check for the exit status missing?

mikrosk:
@th-otto env matrix for targets isn't that simple - see my second point. I'd need to manually enter each multi-target directory and invoke `make $(TRAVIS_TARGET)`. And even that isn't totally trivial because some directories (sys, tools/cops, ...) have very unusual target names.

Tools (or whatever else) can't be really separated because that would require also separated archives. So we wouldn't have freemint-000-123.zip but freemint-, tools- etc for each CPU variant. Very confusing (see how many people are confused _now_).

So that's why I'm thinking this direction - not because I'm bored. :)

th-otto:
> that would require also separated archives. 

So what? Things like cops and hypview are totally independant of the kernel, they just don't belong here. The freemint repo should contain the kernel and drivers, but nothing else. For the same reason that you add those to the archives, you could also add any other app you can think of, like a desktop and editor. Then you end up with a complete distro in a single repo.

You can still provide complete archives like now by downloading the other archives, and merging them in your script. But they shouldn't be build here. That script could even be moved to a separate repo that does nothing else. That would also make it possible to get rid of those OLD_BUILD duplication; you only need one of those scripts here.

>I think the limit isn't 10 000 lines.

It is. That is also documented somewhere on travis. The last line in the log reads "leaving nohog2" directory, but that is not the last action taken.

mikrosk:
Also please bear in mind I don't want want to spend eternity on this task. It's true many things could be cleaned up, improved, unified etc but it doesn't how to be now (and I really don't feel up to it anyway). So I'm looking for a quick(-ish) solution to provide users CPU-based builds with the least amount of work while keeping Travis happy.

Current method of providing 000/020/CF archives is very primitive - the scripts uses `sed` to change the "xxxtargets = ...." line into "xxxtargets = 000 020 v4e" and then invoke plain "make". So anything is better than this, really. My idea/goal/hope is to remove this obscurity, still provide basic 000/020+/CF archives but in addition, if people care, provide also other CPU variants - so not only 020+ but separate 020/030/040/060. That way I could kill OLD_BUILD=1 for good. So to make this work I need some automation.

And btw, yes, adding a desktop and shell will be part of it, too - experience from AF and mailing list clearly shows that people expect the archive to work out of the box.

mikrosk:
Let's discuss OLD_BUILD=1 failing in another issue, same for the CPU= topic. I changed my scripts to work with David's changes, so let's see what breaks. :)

Merging this PR and then my changes in `travis-scripts`.

mikrosk:
And done! I had to fix a couple of minor issues but nothing serious.

If you are wondering what needed to be changed on the travis side: https://github.com/freemint/travis-scripts/commit/9f8d5b80f4b20e563a071cf2d62ecc34676b0a70 (you can also proof-read it if I didn't miss anything).

And the best news is that I have figured out a way how to deploy the env matrix for freemint builds -- using `before_script.sh`. Every instance is going to `sed` the target set to a different value, leading to complete archive for every CPU, making OLD_BUILD=1 obsolete.

Will need some time for that change, though.

Oh and I've changed one more thing -- bintray versions now use 1-xx-yyyyyyyy format because while it's improbable that someone would have more than a few 1-xx-yyy folders in c:\\mint, travis can produce literally hundreds of builds during one month, making the risk of collision far more real.

DavidGZ:
Thanks Miro!

PR #100, opened 2018-08-28
libkern should only depend on CPU type

### LINKED ISSUE #99: libkern's memset broken on MC68000 (alignment issue)
Odd addresses can be only accessed as single bytes (xxx.B instructions) on the MC68000. memset(), as implemented by libkern, tries to account for that: https://github.com/freemint/freemint/blob/02ceb748073b6159ff4bf9929160bd1ed3ed85e1/sys/libkern/asm/memset.S#L24

However, when libkern is compiled for the 68000, the M68000 macro is not defined:

```
m68k-atari-mint-gcc -I../../.. -m68000 -fgnu89-inline -Werror -mshort -g -O2 -fomit-frame-pointer -Wall -Wmissing-prototypes -Wshadow -Wpointer-arith -Waggregate-return -c ../memset.S -o memset.o
```

Hence, code is generated that will crash with an address error on a 68000 when the destination address in odd.

mikrosk:
This is supposed to fix #99.

Not sure about applying $(KERNELDEFAULTDEFS) ? Do we them in libkern?

### DIFF
--- a/sys/libkern/asm/blockcopy.S
+++ b/sys/libkern/asm/blockcopy.S
@@ -13,10 +13,12 @@ __mint_blockcpy:
 	movem.l	4(sp),a0/a1		| dst & src
 	move.l	12(sp),d0		| blocks
 #if defined(__mc68040__) || defined(__mc68060__)
+#ifdef __mc68020__	/* -m68020-60 */
 	move.l	_mcpu,d1
 	cmp.w	#40,d1
 	bcc.s	L_040			| for 68000-68030 do quickmove
 	bra	blkmv
+#endif
 L_040:	dc.l	0xf6218000		| move16 (a1)+,(a0)+
 	dc.l	0xf6218000
 	dc.l	0xf6218000
@@ -54,6 +56,7 @@ L_040:	dc.l	0xf6218000		| move16 (a1)+,(a0)+
 	rts
 #endif
 
+#if !(defined(__mc68040__) || defined(__mc68060__)) || defined(__mc68020__)
 blkmv:
 #ifdef __mcoldfire__
 	lea	-44(sp),sp
@@ -127,3 +130,4 @@ L1:
 #endif
 
 	rts				| return
+#endif	/* !((__mc68040__) || (__mc68060__)) || (__mc68020__) */
--- a/sys/libkern/asm/blockcopy.S
+++ b/sys/libkern/asm/blockcopy.S
@@ -12,7 +12,7 @@
 __mint_blockcpy:
 	movem.l	4(sp),a0/a1		| dst & src
 	move.l	12(sp),d0		| blocks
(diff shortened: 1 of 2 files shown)


th-otto:
Why is this needed? It will only pass flags like PCI_BIOS, NO_RAMFS etc. None of these are used in libkern. If they would, we would be in trouble, because then we would need a separate libkern.a for every kernel target.

mikrosk:
Have you actually tried it? This will pass M68xxx defs only. KERNELDEFAULTDEFS will pass the rest and I haven't included it (hence my question above).

th-otto:
>Have you actually tried it?
Since the patch isn't applied yet: no ;)

But looking at KERNELDEFS:
```
ifeq ($(kernel),ara)
MINT = mintara.prg
CPU  = 040
KERNELDEFS = -DARANYM -DM68040 -DWITH_HOSTFS -DBOOTSTRAPABLE -DC_ONLY
MODULEDIRS = xfs/hostfs
endif
``` 

>KERNELDEFAULTDEFS will pass the rest and I haven't included it 

Both would be wrong. That library should only depend on CPU type. The kerneltargets will link to libkern040.a, not libkernara.a. If you compile for multiple targets that use the same cpu type, you will end up linking to a library for which you don't known which flags were used to compile it, depending on which one is compiled first.

Beside that, grepping through the source, i think there are only 2 files that are affected, memset.S as found by christian, and asm/blockcopy.S.

BTW i think the logic in blockcopy.S is wrong (although harmless). When compiling for 040 or better, you could always use the move16 loop, without checking for cpu type.

mikrosk:
Fair enough - actually now I see that other libkern source also check for the __mc__ defs so perhaps it's an oversight. Both issues are addressed in updated commits.

th-otto:
I think the 2nd one is still not quite right ;) I didn't mean to remove the check for actual cpu type altogether. When you use -m68020-60 that should work also on 020 (-m68020-060 will also define __mc68040__, only when you use -m68040 that check can be omitted).

mikrosk:
Hopefully fine now.

mikrosk:
@th-otto ok to merge?

th-otto:
Oh, wait. While looking at it agaiin, i just saw that you inserted an #endif after the first check. Wouldn't that mean that the code now falls into the move16 case when compiled for plain 68000?

mikrosk:
Isn't that guarded by `#if defined(__mc68040__) || defined(__mc68060__)` ?

mikrosk:
Really not sure what you mean:
diff
On 68000, movem + move would be compiled in and then code starting at `blkmv`, isn't that right?

th-otto:
Ah ok, got a wrong look at the first ifdefs, there are now 2. So go ahead ;)

PR #114, opened 2019-04-28
Call into ACIA interrupt only after posting scancodes to TOS.

czietz:
This makes a modified CKBD (keyboard utility) work with the USB keyboard.

@Perdrix24 reported that CKBD ("Composed Characters Keyboard Driver", https://github.com/ggnkua/Atari_ST_Sources/tree/master/C/Pascal%20Fellerich/COMPOSE/DRIVER/1_5) would not work with the USB keyboard driver. Most of it is better fixed in CKBD itself, which is what I have done. I will release the modified version of CKBD soon.

However, a small change is required to the USB keyboard driver, as well. Since CKBD hooks into the ACIA interrupt, this needs to be called _after_ posting the scancodes to TOS. The call was originally there to be able to reset the Nova screensaver with the USB keyboard. I have verified that moving the call has no impact on the Nova screensaver.

### DIFF
--- a/sys/usb/src.km/udd/keyboard/usb_keyboard.c
+++ b/sys/usb/src.km/udd/keyboard/usb_keyboard.c
@@ -314,8 +314,6 @@ kbd_int (void)
 		kbd_data.newdata.keys[i-2] = 0;
 	}
 
-	fake_hwint();
-
 	/* Handle modifier keys first */
 	DEBUG(("m: %02x", kbd_data.newdata.mod));
 	/* Newly released modifiers */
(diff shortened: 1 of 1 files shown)


czietz:
PS: The Travis CI build error above originates from some problem in setting up the build environment, not from my code change.

mikrosk:
> The Travis CI build error above originates from some problem in setting up the build environment, not from my code change.

Oh, true. I totally forgot about PRs. I had to introduce the [authorization header](https://github.com/freemint/travis-scripts/blob/freemint-master/install.sh#L6) otherwise github API would limit us to 50 requests per hour, what doesn't sound bad unless you realise it is counted per (Travis) IP access. :)

PR #120, opened 2019-05-11
Add debug output support for TOS USB drivers.

czietz:
During USB device driver development, I sometimes need to specifically debug the TOS versions of the drivers. This PR adds TOS support for the DEBUG and ALERT macros used in the code.

Debug output can be enabled with the DEV_DEBUG define, like in the MiNT version of the drivers and will print to the console.

### DIFF
--- a/sys/usb/src.km/global.h
+++ b/sys/usb/src.km/global.h
@@ -80,10 +80,34 @@ typedef char Path[PATH_MAX];
 #undef kfree
 #define kfree Mfree
 
+#ifdef DEV_DEBUG
+
+/* Debug console output for TOS */
+static char tos_debugbuffer[512];
+long _cdecl kvsprintf(char *buf, long buflen, const char *fmt, va_list args) __attribute__((format(printf, 3, 0)));
+static void tos_printmsg(const char *fmt, ...)
+{
+	va_list args;
+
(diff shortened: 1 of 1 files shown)


czietz:
Again, the CI fail is due to some hick-up with Travis, not due to my code.

mikrosk:
> Again, the CI fail is due to some hick-up with Travis, not due to my code.

Yeah, it's because of that public usage of the API which has quite severe per-IP (read: travis virtual IP) limites. I'm not sure how to solve this for PRs (for normal builds we use my token), either download from a static url (i.e. hardcoded release) or not use libcmini at all.

DavidGZ:
Great contribution! I've been missing this feature a lot but never did anything about it. Thanks a lot!

While checking the pull request it has been a surprised for me seeing that TOS drivers are indeed compiled against libkern. Because @atic-atac defined some library functions in [sys/usb/src.km/global.h](https://github.com/freemint/freemint/blob/6e71e4016b9bf7d675f228be1426720191c2c4d3/sys/usb/src.km/global.h#L112) for the TOS drivers I've always thought that TOS drivers didn't have access to the libkern ones. I wonder why @atic-atac redefined some of them.

PR #134, opened 2019-11-29
Make setcookie() fail and not crash if no jar exists.

czietz:
Previously setcookie() would crash with a bus error when no cookie jar existed, i.e., when the pointer to the jar was NULL. This happens on TOS 1.0x when the user does not load anything that creates a cookie jar prior to USB.PRG. 

Now USB.PRG will print an error message instead, like it already did when the jar was full.


czietz:
PS: The build checks fail **not** because of my code but because of issues with the script executed at Travis CI.

PR #136, opened 2019-12-15
Update mint.cnf to reflect default kernel debug level

EmmanuelKasper:
Since in init.c we have
debug_level = TRACE_LEVEL;
reflect this in the documentation.

### DIFF
--- a/doc/examples/mint.cnf
+++ b/doc/examples/mint.cnf
@@ -86,10 +86,12 @@
 # The higher the level, the more stuff MiNT will spew about about
 # what it's doing.
 #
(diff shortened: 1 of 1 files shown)


DavidGZ:
Hi Emmanuel, thank you for pointing this out and the pull request, as you probably have read in the mailing list we have decided to change the kernel's default debug level to ALERT, this has been already done with commit 6f4b394f88d6d2018ca1b88b7ffe0272af65b78a, so this pull request is not necessary any more.

EmmanuelKasper:
@DavidGZ thanks for the  feedback, glad I could help

PR #137, opened 2020-01-10
Remove unused QUANTUM_SMALLER macro from KERNELDEFS 

EmmanuelKasper:
A smaller quantum is anyways defined in sys/mint/mem.h if the M68000 macro is set


mikrosk:
True, it has never ever been used, actually.

PR #142, opened 2020-02-10
Fix _FPU cookie and improve SFP-004 handling

mikrosk:
I was a bit bored so I finally took a look on one bug I had introduced in commit 820a13a0. According to @th-otto's finding in the TOS sources, _FPU 6888x bits are reserved for co-processor mode only: https://mikro.naprvyraz.sk/mint/freemint-list/Week-of-Mon-20170306/000808.html.

So I have finally fixed it (reverted back to original behaviour), i.e. if there is a SFP-004 compatible hardware detected, it sets only bit 16 in `_FPU` cookie value.

However since I love the feature of telling me which FPU have I connected via memory mapped mode, I have extended the original approach so now it properly detects even configurations like MegaSTE + 68881 + PAK68/3 (68030+68882).

Tested on Falcon+68882/CT60, MegaSTE+68881 and MegaSTE+68882.

### DIFF
--- a/sys/arch/detect.S
+++ b/sys/arch/detect.S
@@ -311,22 +311,23 @@ exit:	move.l	a1,sp
 // FPU type detection, experimental (draco@atari.org).
 //
 // This can only detect the hardware FPU, any software emulation
-// will be ignored (a non-zero low word indicates the presence of software
-// floating point support; no specific values have yet been assigned).
+// will be ignored.
 //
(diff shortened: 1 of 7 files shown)


th-otto:
>I was a bit bored

How did *that* happen? ;)

>So I have finally fixed it (reverted back to original behaviour),

Looks good to me.

PR #144, opened 2020-02-12
Fix NVDI/ET4000 handling

### LINKED ISSUE #141: issue between NVDI as graphiccard driver and XaAES
issue between NVDI as graphiccard driver and XaAES.

NVDI Matgraph TC for Matrix TC1006 and TC1208
NVDI Matgraph CXX for Matrix C32 C110 Cards with Intel GPU
NVDI ET4000

The combination do not work. Alternative N.AES 2.0 with MiNT 1.15 works fine. 

See pictures of the destroyed screen ->   https://forum.atari-home.de/index.php/topic,15756.msg

Bootorder are -> INIT.TOS (Matrix Program), REDIRECT.PRG, NVDI.PRG, SLCT_DEV.PRG, MINT.PRG

mikrosk:
NVDI/ET4000 loads additional files by opening the physical workstation from current boot drive. This didn't work as it has been set to U: early on.

Investigation & fix courtesy of Christian Zietz.

Fixes #141.

### DIFF
--- a/xaaes/src.km/cnf_xaaes.c
+++ b/xaaes/src.km/cnf_xaaes.c
@@ -124,7 +124,6 @@ static struct parser_item parser_tab[] =
 	{ "MP_TIMEGAP",            PI_R_S,   & cfg.mouse_packet_timegap },
 	{ "VIDEO",                 PI_R_US,   & cfg.videomode		},
 	{ "ALLOW_SETEXC",          PI_R_S,   & cfg.allow_setexc, Range(0, 2)	},
-	{ "ET4000_HACK",           PI_R_B,   & cfg.et4000_hack	},
 	{ "REDRAW_TIMEOUT",        PI_R_S,   & cfg.redraw_timeout, Range(0, 32000)	},
 	{ "POPUP_TIMEOUT",	       PI_R_S,   & cfg.popup_timeout, Range(0, 32000)	},
 	{ "POPOUT_TIMEOUT",        PI_R_S,   & cfg.popout_timeout, Range(0, 32000)	},
--- a/xaaes/src.km/example.cnf
+++ b/xaaes/src.km/example.cnf
@@ -124,13 +124,6 @@
 # 2: allow all
 #
 #####################################################################
-# et4000_hack (default:0)
-# if XaAES is run with NVDI and an ET4000-graphics-card, it cannot open
-# the physical workstation. Either start from ROM-desktop or open the
-# physical workstation before MiNT by another program. In that case this
-# has to be set.
-#et4000_hack = 1
-#####################################################################
 # xa_bubble (default:0)
 # if you want to use the XaAES-builtin bubble-help (since 1.1.21) set
 # this. You can deactivate any external bubble-app if you use xa_bubble.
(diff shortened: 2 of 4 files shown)


th-otto:
Two small suggestion:

- i would do the change to the boot drive unconditionally. There may be other VDIs suffering from the same problem.

- i would save the current drive, then restore it later instead of unconditionally setting it back to 'U'. The current drive might have been changed in mint.cnf.

mikrosk:
> i would save the current drive, then restore it later instead of unconditionally setting it back to 'U'. The current drive might have been changed in mint.cnf.

I think mint.cnf doesn't apply there because it is `xaloader` which changes it to U:, unless I haven't overlooked something.

th-otto:
But xaloader isn't used when you start N.Aes instead ;)

mikrosk:
But in that case XaAES isn't loaded at all, is it ;-)

th-otto:
That does not mean NVDI is not used. And if i'm not wrong, mint itself already sets the current drive to U. At least i have to insert a "cd c:\" in mint.cnf when running nvdi from it, otherwise it won't start at all.

Generally i think it would be safer to not make any assumptions about the previous current drive.

mikrosk:
Now I'm not sure we are on the same page - I thought the issue was that `xaloader` sets the current drive to U: and then `XaAES` tries to open the physical workstation which in turn calls `NVDI`'s loading routines which fail.

So if the fix itself is located within `XaAES`, whatever I set there shouldn't affect FreeMiNT + N.AES at all, should it?

th-otto:
>whatever I set there shouldn't affect FreeMiNT + N.AES at all, should it?

You never know. If the user had changed the current for whatever reason (maybe because of NVDI, maybe because some other program has similar problems when running from U) then you should not explicitly set it to U after v_opnwk.

For NVDI, there seems to be at least two places where the current drive has to be C (or whatever drive you boot from): when NVDI itself is started (at that time it seems to locate ASSIGN.SYS and the drivers), and when the v_opnwk() call is made.

mikrosk:
Yes but `xaloader` (executed, naturally, immediately before `XaAES` starts) **does** explicitly set it to U: - so whatever happened before that is not important. Once again - we fix this inside `XaAES`, not the kernel.

Also, you can find similar constructions in other parts of XaAES code so that is why I didn't think it is important.

czietz:
@mikrosk: This pull-request looks like what I had in mind when commenting on issue #141. 👍  Just a minor remark about the wording in the comment: This does not only fix NVDI for ET4000 but also NVDI for Matrix graphics cards. Both load files from the current drive.

th-otto:
Well you are right, it is only done in XaAES, so for now it should work. I still think it would be cleaner to save/restore the drive in order not to create other side effects.

BTW, fVDI also tries to locate fvdi.sys in the current path. It just has a fallback to also look in C:\\ if not found there (yes, explicit C:\\, not the boot drive).

mikrosk:
OK, changed the commit wording + let the drive change in all cases, not only for NVDI as I couldn't think of any drawback/negative side effect, as @th-otto has suggested.

Thanks for input guys.

PR #151, opened 2020-03-04
Merge XaAES docs into one HYP file

mikrosk:
Lars Schmidbauer (a.k.a. Lars or Latz) has been brave enough to look at #2 and do something about it.

He has done a pretty nice job with XaAES, you can view his work [here](http://tho-otto.de/hypview/hypview.cgi?url=https%3A%2F%2Fgithub.com%2Fmikrosk%2Ffreemint%2Fraw%2Fxaaes_docs%2Fdoc%2Fxaaes%2Fxaaes.hyp&hideimages=0&hidemenu=0&charset=utf8).

The HYP is part of this PR (for obvious reasons), I can change it to be generated using Thorsten's HCP later + add it as part of the snapshots.

The only drawback here is that Lars has chosen to use .STG as the base and edit/extend/merge that instead of `xaaes.u` because he didn't feel comfortable editing UDO documents & learn its syntax. Personally I find this a missed opportunity (imagine this as a part of a documentation section of the freemint website) but considering the fact he is up to clean up the mint documentation mess next I'm included to approve it as is to encourage him to do further work here. ;)


th-otto:
> he didn't feel comfortable editing UDO documents & learn its syntax

That's a shame, but i understand that. Although it should not be too hard, given that there was already a file (xaaes.u) to start with. I would prefer to convert that back to UDO, mainly because a) we can generate html output of it and b) it makes management of hierarchical structures much easier.

PR #153, opened 2020-03-09
Build scc.xdd for 68000 CPU

### LINKED ISSUE #152: Mega STE scc.xdd fault ...
Mega STE latest freemint 1-19-cur build.

With scc.xdd the kernel hang at boot at the message "SCC serial driver version 0.27". When I disable the scc.xdd the machine boots to the desktop.

czietz:
The MegaSTE has an SCC, too. But scc.xdd was always being built for the 68020-60 and, thus, hung on the MegaSTE. I successfully tested the newly built scc.xdd on my MegaSTE.

Note: Imho, scc.xdd does not warrant per-CPU-builds. Therefore, it's acceptable that now the TT version of FreeMiNT will also get a scc.xdd built for 68000 and upwards.

Fixes #152

### DIFF
--- a/sys/xdd/scc/Makefile
+++ b/sys/xdd/scc/Makefile
@@ -12,7 +12,7 @@ subdir = scc
 
 default: all
 
-CPU = 020-60
+CPU = 000
 
 include $(top_srcdir)/CONFIGVARS
 include $(top_srcdir)/RULES


mikrosk:
Merged, thanks!

PR #157, opened 2020-03-30
Remove redundant GPL2 files

mikrosk:
I think those files are not needed there, they only force the kind of thinking that every binary/module/tool must have a license file because we support so many of them. In fact, everything in FreeMiNT (similarly to Linux) is GPLv2 with the exception of:

- [inet4.xdd](https://github.com/freemint/freemint/blob/master/sys/sockets/COPYING) & many xif modules + [net tools](https://github.com/freemint/freemint/blob/master/tools/net-tools/COPYING) - incredible mess started with Kay Roemer's GPL-incompatible license, then mixed with 3- and 4-clause BSD code and "finished" with GPL headers. I don't think there is a legal way out of it.
- [minixfs](https://github.com/freemint/freemint/blob/master/sys/xfs/minixfs/COPYING) & [minix tools](https://github.com/freemint/freemint/blob/master/tools/minix/COPYING) - a private license by S.N. Henson, IMHO also incompatible with GPL
- [nfsfs](https://github.com/freemint/freemint/blob/master/sys/xfs/nfs/COPYING) & [nfs tools](https://github.com/freemint/freemint/blob/master/tools/nfs/COPYING) - a private license by Ulrich Kuehn, again, IMHO incompatible with GPL

IMHO, if any of those files are mixed with GPL, we are not allowed to distribute binaries, similarly to Linux NVIDIA drivers. Of course I'm not suggesting that. I would suggest just to acknowledge the authors but pretend it's GPL after all because, well... what else we can do?


mikrosk:
Nobody protests, so I'm going ahead. The less clutter the better.

vinriviere:
Indeed, we can say: Everything in FreeMiNT is GPLv2, except special modules, as explicitly specified in their own subdirectories.
And live with the fact that some modules violate the kernel GPL when they are dynamically loaded into it. BTW, is this really true? I'm surprised that the FreeMiNT kernel doesn't allow (as a GPL exception) to load non-free modules. Maybe original MiNT copyright holders weren't aware of such subtlety.

mikrosk:
> BTW, is this really true?

Good question, right? Because as I see it, the only problem here is that the kernel modules include some kernel headers. But what if someone (and I'm pretty sure it has happened) just creates the structure from scratch, say, in asm? He just basically re-implemented the "API".

The biggest worry (if we want to be punctual) is actually that minix, nfs and inet4 code are part of the source tree (they weren't for a long time - so called "mint net" was a separate project), any GPL purist would crucify us for this. ;-)

th-otto:
> some modules violate the kernel GPL when they are dynamically loaded into it

Do they? Isn't that the same as running non-GPL programs in MiNT?

vinriviere:
> > some modules violate the kernel GPL when they are dynamically loaded into it
> 
> Do they? Isn't that the same as running non-GPL programs in MiNT?

Apologies, my above statement is wrong. Nothing happens when you load a non-free module into a Free kernel. For the simple reason that you are allowed to do anything for yourself. Only issues with GPL are about redistribution (unmodified binaries, modified binaries, or derivative work) to someone else.

> Good question, right? Because as I see it, the only problem here is that the kernel modules include some kernel headers. But what if someone (and I'm pretty sure it has happened) just creates the structure from scratch, say, in asm? He just basically re-implemented the "API".

I read somewhere that (roughly) header files couldn't be copyrighted, because they contain only an API, not actual code. I.e. w32api-headers package, part of the MinGW-w64 project.
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/windows.h
https://sourceforge.net/p/mingw-w64/mingw-w64/ci/master/tree/mingw-w64-headers/include/winuser.h
Those files are explicitly in public domain. Even if they contain the same defines as Microsoft headers (same names for defines, structures, functions). However, no Microsoft comments are included there (they would be the property of Microsoft). And this seems to be legal. So any project (GPL or non-free) can use such includes.

Also note that redistributing a GPL executable and a non-free executable in the same archive is also OK. No problem with that. Key point is, if you want to modify the GPL executable for yourself, you must be able to do so. Hence also have the sources for it. The fact that some non-free binary was provided "as a neighbor" in the same archive is not a problem, regarding to the GPL. That could be a problem for big software distributions like Debian which don't accept any non-free software in their main archive. But that's a Debian rule, not a GPL rule.

> The biggest worry (if we want to be punctual) is actually that minix, nfs and inet4 code are part of the source tree (they weren't for a long time - so called "mint net" was a separate project), any GPL purist would crucify us for this. ;-)

Is this a real problem? I don't think that any GPL restriction applies on sources. Rule is: if you get GPL software (i.e. kernel) you are only allowed to redistribute it (either unmodified or modified) as GPL, too. Same applies to any "derivative work", i.e. a heavily modified kernel... or just a single plugin (i.e. module). Basically: when you have 2 different executables, there are never issues, whatever their license is. However, 2 components like shared libraries (or kernel and modules) who share the same process memory space are considered as derivative work.
https://www.gnu.org/licenses/gpl-faq.en.html#GPLPlugins

That being said, what's wrong here?

FreeMiNT should explicitly allow non-free modules. Because it without that, it is forbidden to redistribute non-free binary modules. Forbidden by who? By the FreeMiNT copyright holders. Because such non-free modules, located in the same memory space as the kernel, would violate their kernel license. But are FreeMiNT copyright holders (new and old sources) really going to complain because inetd (another open-source software, different license) is distributed for their kernel? I don't think so.
Adding such exception to FreeeMiNT license would require agreement from all the copyright holders. And as I guess that a lot of them have disappeared, that would be rather complicated to do in a fully legal way.

Let's live with the fact that the status of the FreeMiNT kernel license is not correct regarding to current usages (i.e. non-GPL inetd).

Remember that regarding to licenses, the copyright holder (the source code owner) makes the rules. We could change the license for our own sources. Problem comes with sources borrowed to someone else.

Well, those license issues give headaches.

th-otto:
On Sonntag, 17. Mai 2020 17:21:25 CEST Vincent Rivière wrote:
> ecause such non-free modules, located in the same memory space as the
> kernel, would violate their kernel license.

But wouldnt that also apply to *any* shared library? They are always using the 
same memory space as the process they are loaded into. That would essentially 
forbid any such library to be used.

vinriviere:
> But wouldnt that also apply to *any* shared library? They are always using the same memory space as the process they are loaded into. That would essentially forbid any such library to be used.

GPL is indeed a problem for any shared library. This is why GLIBC uses LGPL so non-free programs can be dynamically (and also statically) linked with it.

Note that (at least on Windows and Linux) there are always a few OS shared libraries in all processes. As a special exception, GPL programs are allowed to be dynamically (or statically) linked with OS libraries. This is a compromise from the FSF to promote Free user programs, even on non-Free operating systems.

PR #163, opened 2020-04-14
add tos_gem and pc_gem widget

ragnar76:
okay, second try


th-otto:
That patch deletes gemflat.rsc & gemflat.hrd. Was that intended?

ragnar76:
yes, it has the same content but the new filename fits better to the pc gem file

th-otto:
Not completely. The WIDG_LOGO is different (a cross in gemflat, now looks the same as a closer)

ragnar76:
ah, yes. i should start writing changelogs :( i don't know why i have changed it but it looks more consistent now

th-otto:
Ok, merged now. Just renamed the hrd file to lowercase.

PR #187, opened 2020-06-07
XaAES debug fixes

chrisridd:
Diagnostic logging in XaAES seems to have bit rotted slightly. With these changes I can now start a desktop, do simple things, and shutdown.

### DIFF
--- a/xaaes/src.km/xa_rsrc.c
+++ b/xaaes/src.km/xa_rsrc.c
@@ -462,12 +462,13 @@ fix_tedarray(struct xa_client *client, void *b, TEDINFO *ti, unsigned long n, ch
 			ti->te_ptext	= (char *)-1L;
 			ti->te_ptmplt = (char *)ei;
 
-			ti++;
-			ei++;
-			n--;
 			DIAG((D_rsrc, NULL, "fix_tedarray: ti=%lx, ptext='%s'", (unsigned long)ti, ei->ti.te_ptext));
 			DIAG((D_rsrc, NULL, "ptext=%lx, ptmpl=%lx, pvalid=%lx",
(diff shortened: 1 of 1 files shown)


th-otto:
Your 2nd patch generated warnings when compiled without GENERATE_DIAGS, so i modified it slightly to print the messages at the start of the loop instead.

chrisridd:
Sure, your mods look OK too. Thanks! Hopefully this is the last bit of yak shaving before I can investigate my Lattice issue...

th-otto:
To investigate that issue, you should maybe use single debug printfs in the area you want to investigate, instead of generating lots of useless diag output. Just an idea, but i would try to trace calls to wind_update(),i suspect that there might be some lock that is not released again.

DavidGZ:
@chrisridd to do what @th-otto is suggesting I use the function FORCE(), once you know what piece of code must be investigated this will allow you print debug messages without turning on the debug messages for the rest of the code, also works independently of your setting in MiNT init menu.

chrisridd:
> On 7 Jun 2020, at 09:22, DavidGZ <notifications@github.com> wrote:
> 
> 
> @chrisridd to do what @th-otto is suggesting I use the function FORCE(), once you know what piece of code must be investigated this will allow you print debug messages without turning on the debug messages for the rest of the code, also works independently of your setting in MiNT init menu.

Thanks for the tip about FORCE() - I think watching wind_update might indeed be interesting but as at the moment I’m just not sure what Lattice’s custom form_do() is doing, I’ll need to start with a bit of a firehose...

It is a shame that strace cannot trace AES calls made by a process. That’d be quite cool.

PR #203, opened 2020-09-28
USB: Use unique wakeup condition for every hub

### LINKED ISSUE #202: FreeMint 1-19-e2521230 w/ LightningVME driver Hubpoll CPU usage
Dear Freemint team,

I have been in contact with the developers of LightningVME about abnormally high CPU usage in FreeMint in conjunction with the LightningVME drivers. 
The problem as I see it: The CPU usage of the 'hubpoll' processes steadily increases after boot, going from 5-10%, up to 30% after a minute. I have extensively tested the driver with the developer and he suggested we open an issue here, as he feels it may not be down to the driver. He also sent me an updated driver for testing which made no difference.
Nothing plugged in to the USB:
nothing_in
Just 1x Logitech USB2 4 port hub:
justhub
Logitech hub & USB Asix LAN plugged in:
allin_hub_and_asix

As you can imagine this really slows down the machine during usage. I am using very latest FreeMint snapshot and Lightning VME drivers. Machine is a TT with 64mb fastram and NOVA mach64. Res 1024x768 256c.

czietz:
Previously, the timeout function installed by one hubpoll thread
would wake up all hubpoll threads, causing a high CPU load.

This fixes #202.

### DIFF
--- a/sys/usb/src.km/hub.c
+++ b/sys/usb/src.km/hub.c
@@ -598,19 +598,26 @@ usb_hub_events(struct usb_hub_device *hub)
 void
 usb_hub_poll(PROC *p, long device)
 {
-	wake(WAIT_Q, (long)&usb_hub_poll_thread);
+	/* Device address is used as wakeup condition. */
+	wake(WAIT_Q, device);
 }
 
 void 
 usb_hub_poll_thread(void *ptr)
 {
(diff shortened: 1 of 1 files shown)


DavidGZ:
Thanks!

PR #207, opened 2020-10-12
Update french-pc.src

Landemarre:
Add AltGR code for non Eiffel adapter (Useful for Aranym and V4SA)


vinriviere:
In 2013 I created this table for my FireBee with an Eiffel adapter and a PS/2 keyboard. IIRC, it worked perfectly. Does this mean that Eiffel and ARAnyM don't generate the same scancode for the AltGr key? Is there any standard for that?

Landemarre:
Hello Vincent

I think with Aranym option AltGR = Yes I have same as physical keyboard, 
I don't know if I remove this option.

My Understand Eiffel help with system with no AltGR table and use for 
this I think Alternate  key and need more key code to not be in conflict 
with standard key.

As it was define the AltGR table in all case was not use, I think with 
eiffel it use Alt table or alt caps table, but under Mint I don't know 
how it could work with Emutos as system boot as Alt and AltCaps was not 
loaded (to load it need tosver 4 mini), except to launch a TSR with Mint 
to fix this.

This is my understand.

I do a change in keyboard.c to load in all case all table should so Alt 
table too, so should work with Eiffel

Olivier

Landemarre:
I confirm Aranym not use Eiffel key code even with AltGR = no

PR #208, opened 2020-10-12
Keyboard patch all in one

Landemarre:
Full keyboard table support on any system.

To be able support AltGR on system with tosver < 0x400, useful under Aranym and V4SA as they not use Eiffel extension
Add key for AltGR table in french PC keyboard


mikrosk:
I squashed your changes into one commit (so the clutter with merging other PRs is not visible), for the next time I recommend to look up so called "cherry-pick", it will make your life easier in such situations ;-) (you'd create a new branch, cherry-pick those three commits, done!)

PR #209, opened 2020-11-09
USB: storage: Ignore REPORT_LUN command

czietz:
Not all USB mass storage devices support this, some crash rather
badly. HDDRIVER's HDDRUTIL program sends this command, so it's
blocked at the driver level. Note that this does not restrict
the drivers ability to handle multiple-LUN devices such as card
readers.

This was a regression introduced with commit c176b37363f6bc6df38b400d0828bd482a338d6d.

### DIFF
--- a/sys/usb/src.km/udd/storage/usb_scsidrv.c
+++ b/sys/usb/src.km/udd/storage/usb_scsidrv.c
@@ -293,6 +293,10 @@ SCSIDRV_In (SCSICMD *parms)
 				retries = 10;
 			}
 
+			if (srb.cmd[0] == SCSI_REPORT_LUN) {
+				return -1;
+			}
+
 			/* promote read6 to read10 */
 			if (srb.cmd[0] == SCSI_READ6)
 			{


czietz:
Additional information: This comes from a Lightning customer complaint, where it was mentioned that a USB stick was not recognized by HDDRUTIL. With said USB stick, I was able to reproduce the issue and I could trace it back to the REPORT LUNS command that HDDRUTIL sends. Unfortunately, the controller on this particular USB stick completely crashes when receiving the command; no further communication is possible.

References that REPORT LUNS is a problematic command for USB storage devices can be found elsewhere, e.g., in the Linux kernel: https://github.com/torvalds/linux/blob/f8394f232b1eab649ce2df5c5f15b0e528c92091/drivers/usb/storage/scsiglue.c#L341-L347

As already mentioned in the commit message, blocking this command does not impede HDDRUTIL from detecting devices with multiple LUNs as it will simply try to enumerate them.

DavidGZ:
Thanks for the investigation and the fix

Issue #219, opened 2021-05-06
Regression: Commit 5e8dd40e ("USB:storage: Support write protection") makes certain USB sticks hang


czietz:
Unfortunately, commit https://github.com/freemint/freemint/commit/5e8dd40e53b6535188f9bcf956d4ff770ccd8a0c causes quite a major regression: several USB sticks (unbranded, possibly using the same controller chip) now hang indefinitely when connected to the Atari, making them (and the whole system) unusable.

I have confirmed on another OS that in fact the MODE SENSE 10 command introduced by this commit crashes the firmware on these USB sticks. Therefore, this is _not_ an issue with the USB stack (or the Lightning hardware), but an inherent limitation of these USB sticks.

I'm afraid that I currently have no proposal (other than to revert the commit) to fix this issue. I am wondering though what this commit is even supposed to achieve. It queries the write protection status (via MODE SENSE 10), but I don't see that this information is even used anywhere. Can someone please explain? If the write protect status is not used by the driver, why is it queried?

PS: Tagging @DavidGZ @Perdrix24

czietz:
PS: One of the affected sticks is the one that was also the cause for https://github.com/freemint/freemint/pull/209. One can assume that some USB stick controller firmware is quite bug-ridden. 

Unfortunately, users still say "but it works with Windows/Linux/macOS/...". Thus, imho, the USB stack must be able to work even with such devices.

Perdrix24:
> Unfortunately, commit [5e8dd40](https://github.com/freemint/freemint/commit/5e8dd40e53b6535188f9bcf956d4ff770ccd8a0c) causes quite a major regression: several USB sticks (unbranded, possibly using the same controller chip) now hang indefinitely when connected to the Atari, making them (and the whole system) unusable.
> 
> I have confirmed on another OS that in fact the MODE SENSE 10 command introduced by this commit crashes the firmware on these USB sticks. Therefore, this is _not_ an issue with the USB stack (or the Lightning hardware), but an inherent limitation of these USB sticks.
> 
> I'm afraid that I currently have no proposal (other than to revert the commit) to fix this issue. I am wondering though what this commit is even supposed to achieve. It queries the write protection status (via MODE SENSE 10), but I don't see that this information is even used anywhere. Can someone please explain? If the write protect status is not used by the driver, why is it queried?

[...]

PR #220, opened 2021-05-07
USB:storage: For write protection query use MODE SENSE (10) or (6)

czietz:
Not all USB mass storage devices support the MODE SENSE (10) SCSI command.
In particular, the firmware of certain USB sticks even crashes when receiving
this command. Thus, as also done by Linux, the subclass of a device decides
whether to use MODE SENSE (10) or MODE SENSE (6).

This fixes https://github.com/freemint/freemint/issues/219


DavidGZ:
Thanks for the fix and the investigation!

Issue #225, opened 2021-06-30
Purpose of minthat.prg


czietz:
Because of [this forum thread](https://www.atari-forum.com/viewtopic.php?f=126&t=40663) I was wondering what purpose _minthat.prg_ (included in snapshot builds) still serves. Yes, it is built with `OLDTOSFS` enabled:
https://github.com/freemint/freemint/blob/b91b7c5d856bb7d8e5205d97ab185fe25de6283e/sys/KERNELDEFS#L122
... but it does not work with Hatari's GEMDOS drive emulation, either:
![grab0001](https://user-images.githubusercontent.com/9830685/124001655-252e8800-d9d5-11eb-9ffa-69f00319695a.png)
(Tested by manually renaming _minthat.prg_ to _mint000.prg_ so that it gets loaded.)

And when using an ACSI/IDE/SCSI HD image, one can use the regular FreeMiNT kernels under Hatari. Which is the use case  that _minthat.prg_  serves?

DavidGZ:
This thread have some background about this issue:

https://mikro.naprvyraz.sk/mint/201502/msg00114.html

czietz:
Maybe I missed the crucial detail in that mailing list thread, but the purpose of _minthat.prg_ is still not clear to me.

mikrosk:
I think the moral story of that thread was that the OLDFS was supposed to work and for some reason it doesn't anymore.

czietz:
Well, I seem to have identified and fixed the issue. This is _minthat.prg_ with OLDTOSFS running from a GEMDOS emulated drive C: in Hatari:

![grab0001](https://user-images.githubusercontent.com/9830685/124156212-a2bacc80-da97-11eb-9e0e-2428953942b1.png)

Expect a patch / pull request as soon as I have some time for it.

czietz:
Confirmed to be working now with the snapshot build of https://github.com/freemint/freemint/commit/b44c975ecbc4a36cc2d90251622e24650c2852d4.

PR #226, opened 2021-07-01
Make OLDTOSFS work with Hatari's GEMDOS drive emulation

czietz:
As mentioned in #225, the special _MINTHAT.PRG_ imho only serves a purpose if it works with Hatari's GEMDOS drive emulation. As there is no BIOS device, media change detection and, thus, ultimately every drive access would fail with ENODEV (-15). (This can be seen in my first screenshot posted in #225.) 

Note that a similar change was introduced by Helmut K via commit 44341fef3869c45d805964ccbe2bae7204951582 but later reverted with commit 2dfde1b15202fca5f4c6ceeacce5cec6087b17d0.

### DIFF
--- a/sys/filesys.c
+++ b/sys/filesys.c
@@ -678,7 +678,16 @@ disk_changed (ushort d)
 		 * Note that ENODEV must be tested for drives A-C, or else
 		 * booting may not work properly.
 		 */
+#ifdef OLDTOSFS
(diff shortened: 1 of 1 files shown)


DavidGZ:
Thanks for the fix

th-otto:
Yes, thanks for the fix. I just wonder whether that should also be done without OLDTOSFS (why should we continue to check for a drivechange, when we get ENODEV, even for drive C?)

PR #232, opened 2021-10-02
Fix broken palette remapping in 8bpp mode

### LINKED ISSUE #171: XaAES always remaps textures into the default palette
Joakim reported a very old (and still present) bug of XaAES, along with a pretty accurate analysis what is most likely going on:

> I can't help but noticing that I have the same palette
> problem (perhaps limited to 256 colours, since that is the only thing
> I am running here) with XaAES current version (daily build binaries
> from 2020-05-03)  as with my previously installed 2013 build of
> 1-19-cur. The problem is that the IMG textures for the GEM windows of
> XaAES are converted to wrong palette if NVDI is set in XAAES.CNF
> "palette = <name>". The IMG textures come out with colours all garbled
> up.
> 
> Since I am pretty sure that these IMG files are remapped by XaAES when
> it is launched, I am starting to suspect that XaAES is setting the
> palette specified in XAES.CNF as per "palette = <name>" but only
> *after* the IMG files have already been remapped.
> 
> So in essence, I think that the IMG files are remapped to the old
> Atari GEM palette (which seems to be the one present when XaAES is
> launching), and when XaAES later is issuing the switch to the NVDI
> palette, the IMG files are doomed to be looking wrong. And since
> texture image files are loaded/remapped during XAES startup sequence,
> it does not seem like they are affected by any other palette related
> entries to the xaaes config, like e.g. icn_pal_name. At least, nothing
> I tried editing seems to make a difference.

mikrosk:
Fixes #171.

That whole file is a huge mess so I'm just hoping this doesn't break something else. fVDI @ 16-bit seems to be still working correctly, though.

### DIFF
--- a/xaaes/src.km/trnfm.c
+++ b/xaaes/src.km/trnfm.c
@@ -184,11 +184,11 @@ static const short devtovdi8[256] =
 /*                                 0   1   2   3   4   5   6   7   8   9   0   1   2   3   4   5 */
 static const short devtovdi4[] = { 0,  2,  3,  6,  4,  7,  5,  8,  9, 10, 11, 14, 12, 15, 13,  1 };
 
-#if 0
 /*                                                                         1   1   1   1   1   1 */
 /*                                 0   1   2   3   4   5   6   7   8   9   0   1   2   3   4   5 */
 static const short vditodev4[] = { 0,255,  1,  2,  4,  6,  3,  5,  7,  8,  9, 10, 12, 14, 11, 13 };
 
+#if 0
 static const short vditodev4[] = { 0,  1,  2,  3,  4,  5,  6,  7,  8,  9,  8,  9, 12, 14, 11, 13 };
 #endif
 
@@ -622,13 +622,12 @@ build_pal_xref(struct rgb_1000 *src_palette, struct rgb_1000 *dst_palette, unsig
 				c = j;
 			}
 		}
-		if( screen.planes == 8 )
+		if (c < 16 && screen.planes == 8)
 		{
-			/* ?? */
-			if (c == 9)
-				c = 8;
-			if( c == 1 )
-				c = 255;
+			/* We need to convert from vdi to dev here so vr_trnfm() converts
+			 * back to what we already know here...
+			 */
+			c = vditodev4[c];
 		}
 
 		cref[i] = c;
--- a/xaaes/src.km/trnfm.c
+++ b/xaaes/src.km/trnfm.c
@@ -180,20 +180,21 @@ static const short devtovdi8[256] =
 #define G_LYELLOW		14
 #define G_LMAGENTA		15
 #endif
-/*                                                  1  1  1  1  1  1 */
-/*                           0  1  2 3 4 5 6 7 8  9 0  1  2  3  4  5 */
-static const short devtovdi4[] = { 0, 2 ,3,6,4,7,5,8,9,10,11,14,12,15,13,1  };
-
-/*                           0   1  2 3 4 5 6 7 8  9  0  1  2  3  4  5 */
-/*static const short vditodev4[] = { 0,255, 1,2,4,6,3,5,7, 8, 9,10,12,14,11,13 };*/
+/*                                                                         1   1   1   1   1   1 */
+/*                                 0   1   2   3   4   5   6   7   8   9   0   1   2   3   4   5 */
+static const short devtovdi4[] = { 0,  2,  3,  6,  4,  7,  5,  8,  9, 10, 11, 14, 12, 15, 13,  1 };
 
 #if 0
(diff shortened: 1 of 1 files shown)


th-otto:
That looks strange to me. In build_pal_xref, "c" is a pixel index. But you convert it using the vdi to pixel lookup table.

But i have to admit i don't understand the original code either. Converting pixel index 15 to 255 would make sense, but nothing else.

mikrosk:
I think that table is still palette indexes (as it contains only entries 0 - 255). Another thing which makes me believe this is correct is the code which got replaced -- those two super-strange `if`s are basically two entries from that table, i.e. for some reason Helmut thought the conversion is not needed and then he re-added those two indexes anyway (with others left as they are).

th-otto:
Yes, but the palettes form src and dst are indexed by pixel index, as well the cross-reference table he is building. VDI->pixel conversion is a different thing.

th-otto:
Hm, ok, or maybe not. In remap_bitmap_colindex, he does another lookup through the dev->vdi table. That is IMHO nonsense, but has the effect that he essentially uses the VDI index as lookup into cref, so in that case your fix would be correct. Since remap_bitmap_colindex is also used in a few other places, i don't think this should be changed currently, but that whole thing is still a big mess. Apart from that, the remap function is far from being optimal.

That thing has still to be carefully checked. remap_bitmap_colindex is also called with md == 0 in some places, in which case that vdi to pixel lookup is not done, and then build_pal_xref  must also not do that transformation.

GokMasE:
I am really glad to see this being looked into, I did have thedistinct
feeling that some hasty work was what introduced this problem. Around
that time, the keyboard navigation also had some "improvements" done
to it I believe. Is the suggested fix for the palette problem included
in the automated bootable build from 2021-10-01 18:46 by any chance?

If so I'd of course be keen to see how that performs here - with a
fairly recent MiNT kernel present I presume it would suffice to just
replace XaAES?


On 03/10/2021, Thorsten Otto ***@***.***> wrote:
> Hm, ok, or maybe not. In remap_bitmap_colindex, he does another lookup
> through the dev->vdi table. That is IMHO nonsense, but has the effect that
> he essentially uses the VDI index as lookup into cref, so in that case your
> fix would be correct. Since remap_bitmap_colindex is also used in a few
> other places, i don't think this should be changed currently, but that whole
> thing is still a big mess. Apart from that, the remap function is far from
> being optimal.
>
> That thing has still to be carefully checked. remap_bitmap_colindex is also
> called with md == 0 in some places, in which case that vdi to pixel lookup
> is not done, and then build_pal_xref  must also not do that transformation.
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/freemint/freemint/pull/232#issuecomment-932888418


-- 
Joakim Högberg
Borevägen 30
941 42  Piteå

th-otto:
Currently this is only a pull-request which has not been merged yet, so these changes are not available as snapshot.

mikrosk:
The biggest trouble for me here is the fact that my knowledge of AES/VDI is very limited. Until now, I didn't even have a slightest idea that there's some dev/vdi mapping needed and what it actually means.

So I could only "reverse engineer" the original change and follow code flow. If you can Thorsten make heads or tails out of it, I'm all in for a better fix.

Ideally, I would just delete the whole file and start from scratch, there's so many hacks and spaghetti code that it's really hard to follow / debug...

th-otto:
The mapping is needed because VDI uses different indices to specify colors than is written to screen memory. The tables are also in TOS hyp: https://freemint.github.io/tos.hyp/en/VDI_fundamentals.html#CLUT  (the table for 256 colors is almost identical to the one for 16 colors, except that VDI color 1 is pixel index 255)

As a result, when you specify vs_color(GREEN,..) for example, you are actually modifying the color for pixel index 2 etc.
The idea is rather simple: in your application, you always use vst_color(GREEN) etc, and that should work in all resolutions. Also note that BLACK (VDI index 1) is always mapped to the highest color number available, so pixels drawn in black always have all bits set, while pixels drawn in in WHITE always have all bits clear (except in hicolor/truecolor modes, where it is the opposite).

As already said i above, i currently don't understand why there is a mapping from VDI to pixel indices in the remap_bitmap_colindex function. That does not make much sense to me. For my understanding, that should just change the pixels in the src image to match the palette actually in use (which is needed, because just cannot change the whole palette every time just because some window needs a different palette than another). But as always, there *might* be a reason for doing so, just it is nowhere mentioned in the comments.

Starting from scratch may be a bit of overkill, but looking at the code it may actually be needed. Overall, it seems to be much more complicated than necessary, and could also need some optimizations. When remapping for example, i think you only have to deal with conversion from 16->256 colors and 256->256 colors(with different palettes) . That should eliminate the need for the innermost loop.

Issue #22, opened 2017-02-13
XaAES crashes with memory protection without NVDI


mikrosk:
IIRC, in the b/w mode XaAES works fine as is. However in >=4 (?) colours it requires NVDI/fVDI to run. I think it shouldn't be too hard to detect this situation and warn the user appropriately.

skaftetryne:
XaAES is working fine on my Milan without NVDI or fVDI in 8- and 16-bit modes. It is also working fine (but slow!) on my Mega ST with VOFA graphics card without NVDI or fVDI in 8-bit mode.

mikrosk:
Interesting. Maybe it causes trouble only in EmuTOS?

skaftetryne:
I'm sure I've had problems with XaAES on my standard Falcon with 4.04 and no graphics card. So there is a problem, just not sure exactly what.

mikrosk:
Come to think about it, maybe the planar modes are the problem? I'll do some testing when I find time.

skaftetryne:
Dug out my stock Falcon last night for a little bit of testing.

- XaAES works fine with the TOS VDI in mono.
- In colour accessories are loaded, but there's a bus error when apps (like the desktop) is loaded. Could not see any graphical/visual glitches.

On my STE with 8Mb alt-RAM and ET4000 graphics card:

- Works fine in both mono and colour on the graphics card with just Nova VDI.
- Works fine in mono on the shifter (could not test colour as I don't have a colour monitor at hand now), but only if alt-RAM is disabled. If alt-RAM is enabled and NVDI is not running, XaAES will crash upon launch. That is probably caused by blitting to/from alt-RAM.

zabruder:
I don't think it's an XaAES-bug. I can reproduce it in aranym-mmu in 8 bit video. Applications crash in v_opnvwk (PC: 00E33BDA). Sometimes, sometimes not.  I can also reproduce it with TOS 3.06 on the TT with the TOS-VDI but haven't traced it there. Maybe the (Emu)TOS-VDI makes assumptions about the (Emu)TOS-AES opening the physical workstation or maybe it's just a stack-overflow.

vinriviere:
I vaguely remember that long ago, XaAES crashed with EmuTOS VDI. This was because it calls AES appl_init() from supervisor mode. While this is generally considered as a bad thing, it should actually work fine. But this leaded to a stack overflow, and we finally added a workaround. I don't remember well, maybe the stack overflow problem appears again?

[...]

Issue #116, opened 2019-05-04
trap #1 handler always reads 36 bytes from stack


mikrosk:
I found this totally by accident while looking for something different in my gmail archive. There has been an issue of crashes/memory violations during early startup.

@vinriviere had investigated this thoroughly in this thread: https://mikro.naprvyraz.sk/mint/200906/msg00039.html and provided a workaround for TOS4 and (then assumed) earlier FreeMiNT kernels: https://github.com/freemint/mintlib/commit/731463390476e37c97cb5a3e407316820164bf0c.

However @atic-atac has never fixed the issue in the kernel, as he wrote here: https://mikro.naprvyraz.sk/mint/200906/msg00015.html so even though all mintlib-compiled binaries are fine, the bug is still there: https://github.com/freemint/freemint/blob/master/sys/arch/syscall.S#L166. The handler shouldn't blindly copy the stack space because there may not be any parameters pushed to the stack and reading from forbidden memory.

This isn't a high-priority issue but it is nice to have it written down I think.

th-otto:
The fix is already there, just not active. When defining NEW_SYSCALL,  the handler looks in a table how many words need to be copied for each call.

Also, programs using mintlib aren't affected by this, because there is a workaround by initializing the SP 64 bytes less than would theoretically be possible.

I guess it will be hard to find a configuration which actually triggers the bug. First off, the memory is only read, so that can only be a problem with memory protection enabled. Then, it must be a GEMDOS call that is done rather early in the program, because later the SP will most likely be already pushed down enough. And lastly, within the 36 bytes that are copied you have to cross a page boundary.

mikrosk:
Literally everything you have mentioned has been discussed/reproduced in that thread :) The trouble with NEW_SYSCALL is that that isn't the fix, it was there before, too. In other words, nobody knows in what shape it is, how experimental it is etc.

vinriviere:
I don't remember well, but if I'm not wrong this can easily be triggered with a simple hello world written in assembler, on a machine with Alt-RAM. If the program has the altram flag in the header, it will be loaded to the biggest block of free Alt-RAM, which is the whole TTRAM by default. The OS sets the process stack at the very top less 8 bytes (with basepage and 0 already pushed). Whenever you call a trap (i.e. Cconws), implementation must only read the pushed parameters, no more, otherwise read will happen after the end of TTRAM and it will cause a Bus Error.
[...]

PR #251, opened 2022-04-09
Fix TOS 3/4 VDI with memory protection

mikrosk:
This is my take on #22 . There were two separate issues with it:
- syscall context was no longer saved as expected: since the introduction of reentrant syscalls the PC in the `ctxt` struct wouldn't belong to the last caller (TOS VDI routine in our case) but to the user application calling `v_opnvwk` (i.e. the malloc() call would be made while [the in_kernel flag](https://github.com/freemint/freemint/blob/master/sys/arch/syscall.S#L699) is already true and not saving any context)
- xaaes.km is a kernel module and therefore all functions (incl. [its v_opnvwk](https://github.com/freemint/freemint/blob/master/xaaes/src.km/k_init.c#L726)) are always called with `in_kernel` flag set so no new context is built/saved either.

As you can see, I have solved it via an addition flag, `in_vdi`, which signalises whether the currently processed syscall is originating from a VDI call (trap 2 call to be precise but AES has its own dispatcher in the kernel so it's basically just VDI all along).

Additionally I have introduced a flag `tos_vdi` in kernel which ignores this workaround in case of NVDI/fVDI/NOVA VDI/EmuTOS VDI has been found.

Btw that `syscall.S` file could use some cleanup. We should decide what to do especially with the `NEW_SYSCALL` code, it makes everything super hard to read and follow.

### DIFF
--- a/sys/Makefile.objs
+++ b/sys/Makefile.objs
@@ -25,7 +25,7 @@ WARN =  -Wall \
 	-Wpointer-arith \
 	-Wcast-qual
 LDFLAGS = -nostdlib
-LDFLAGS += -Wl,-Map -Wl,map.txt
+LDFLAGS += -Wl,-Map=map.txt
 # XXX should be come from arch/
 LDFLAGS += -Wl,--entry -Wl,_main
 LIBS = $(LIBKERN) -lgcc
@@ -63,7 +63,7 @@ VPATH = ..
 build: $(MINT)
 
 $(MINT): $(OBJS) $(BUILDINFO) $(LIBKERNTARGET)
-	$(CC) $(CFLAGS) $(LDFLAGS) -s -o $@ -Wl,-Map=mint.map \
+	$(CC) $(CFLAGS) $(LDFLAGS) -o $@ \
 		$(shell for file in `cat $(srcdir)/arch/$(compile_dir)/startup` ; \
 			do echo $(srcdir)/arch/$(compile_dir)/$$file ; done) \
 		$(OBJS) $(BUILDINFO) \
--- a/sys/arch/kernel.S
+++ b/sys/arch/kernel.S
@@ -17,6 +17,7 @@
 #include "magic/magic.i"
 
 	.globl	_enter_gemdos
+	.globl	_enter_vdi
 	.globl	_enter_kernel
 	.globl	_leave_kernel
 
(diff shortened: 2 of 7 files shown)


th-otto:
Just a few comments:

- why did you remove the -s flag in the Makefile? IMHO a kernel with symbols is rather useless, since you won't be able to run it in a debugger.
- The check for NVDI/fVDI won't work. Both are started either from mint.cnf, or after mint, so their cookies are not yet set.

mikrosk:
> why did you remove the -s flag in the Makefile? IMHO a kernel with symbols is rather useless, since you won't be able to run it in a debugger.

Hatari debugger, anyone? ;) It has been a huge help when working on this issue it so it is a real world scenario (while you can strip the binary with `make strip`, you can "unstrip" it with this flag in place)

> The check for NVDI/fVDI won't work. Both are started either from mint.cnf, or after mint, so their cookies are not yet set.

Good catch! I wanted to optimise it a little (so the check wouldn't be done for every M(x)alloc call). I'll change it.

> Previously, without saving the context for the Malloc, the pc was checked for being in ROM. This was the PC the v_opnvwk call, and therefor exactly what was needed. I quite don't understand why the patch should fix the problem (although is a bit cleaner that checking for absolute ROM addresses).

It's because of that re-entrancy. Before this feature had been implemented by Ozk, all calls were quite straightforward -- an application makes a syscall, we store PC of the caller. If it's chained (`v_opnvwk` -> `Malloc`), the caller is ROM because this is where `v_opnvwk` would be implemented. However now the kernel stores the application context by the first call (`v_opnvwk`) and by the next call (`Malloc`) kernel directly jumps to its re-entrant handler, without storing anything. So `ctxt` still contains the application PC and not ROM PC. And for `xaaes.km` it is even more complicated, see above.

> I agree with the NEW_SYSCALL define. But that would probably mean we have to remove it, since that code was never tested.

I guess it was tested by Konrad back in the 2000s but its introduction by commit 54a7481f doesn't tell much about its purpose.

th-otto:
> So ctxt still contains the application PC and not ROM PC.

Yes, but that is intended in this case. You want to catch the v_opnvwk case, not the Malloc case.

mikrosk:
> Yes, but that is intended in this case. You want to catch the v_opnvwk case, not the Malloc case.

Not sure what you mean? `v_opnvwk` is a normal application call, so naturally its origin is from application memory. On the contrary, we want to catch the `Malloc` case which is called from ROM's `v_opnvwk` implementation.

th-otto:
About the NEW_SYSCALL: the handler copies the trap parameters to the kernel stack. currently it does that blindly by copying 32 bytes, with NEW_SYSCALL only the actual needed parameters are copied. That may prevent problems with MP when SP+32 is maybe outside of the application stack. But i don't think it can be cleanly done that way, there are calls like VSetscreen/Rwabs that take extra parameters only in certain cases, and there may also be resident programs that install new functions which the kernel does not even know about.

mikrosk:
Ah, right, linking #116 for reference.

mikrosk:
I have implemented a few optimisations (like completely disabling this mechanism on plain 68K) and also simplified the condition to avoid unnecessary `get_cookie` calls. I was tempted to remove the check completely (i.e. rely on VDI to supply a non-zero `mode`) but then I realised that VDI could make another M(x)alloc call which would be completely fine to have a private memory and we would force it to be 'super accessible' for no reason.

PR #253, opened 2022-04-23
Introduce hat030 and replace ARANYM with WITH_NATIVE_FEATURES

mikrosk:
I've realised that more than often I need this target for some advanced FreeMiNT debugging in Hatari. It offers natfeats and GEMDOS emulation (and removes natfeats from regular 030 builds introduced in 3661519b).


mikrosk:
I've reworked this one after realising the duplicity and unreliability of the original commit https://github.com/freemint/freemint/commit/3661519b9cfbb0fc0498e75343ca734ac42a0986 (one had to have `__NF` cookie installed, i.e. it wouldn't work on Hatari + TOS)

I have also added `WITH_NATIVE_FEATURES` to the 68000 Hatari build because it increases file size only by a couple of KBs. Interestingly, it is `OLDTOSFS` which increases 68000 binary size by 32 (!) KB and 68030 one by 10 KB.

@th-otto do you see any issue with this change?

PR #270, opened 2022-07-13
Remove C_ONLY code

mikrosk:
It has been used only for ColdFire and Aranym targets for no apparent reason.

@mfro0 can you imagine a reason why the C code would be useful for the FireBee?

### DIFF
--- a/sys/KERNELDEFS
+++ b/sys/KERNELDEFS
@@ -6,7 +6,6 @@
 # -DCRYPTO_CODE		include the cryptographic filesystem stuff
 # -DSOFT_UNITABLE	loadable unicode tables
 # -DBUILTIN_SHELL	include the built-in minimal shell
-# -DC_ONLY		replace some assembly function by C compatible one
 # -DWITH_MMU_SUPPORT	include MMU support
 # -DPCI_BIOS		include PCI_BIOS interface emulation
 
@@ -99,7 +98,7 @@ endif
 ifeq ($(kernel),col)
 MINT = mintv4e.prg
 CPU  = v4e
-KERNELDEFS = -DNO_FAKE_SUPER -DC_ONLY -DPCI_BIOS
+KERNELDEFS = -DNO_FAKE_SUPER -DPCI_BIOS
 kernel_nocflags = -DWITH_MMU_SUPPORT
 endif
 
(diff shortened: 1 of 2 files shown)


czietz:
I don't know. But I like to guess: The _other_ implementation (i.e., without the C_ONLY define) calls an assembler function `quickzero`. Back when the C_ONLY code was added, this assembler function (see https://github.com/freemint/freemint/tree/8d19d411b2b90b7f32f7af247f9506138cc28253/sys/libkern/asm) was not yet ported to ColdFire. I.e., it used instructions not available on ColdFire.

Landemarre:
> It has been used only for ColdFire and Aranym targets for no apparent reason.
> 
> @mfro0 can you imagine a reason why the C code would be useful for the FireBee?
> 
> Or maybe @Landemarre remembers why he added it back in 2007?

Hello

You are true, I add this I remember very well, it was when I port Mint from GCC 2.95 to GCC 4 for coldfire reason, as I have absolutely no knowledge in coldfire ASM, I want have the minimal possible pure 68000 ASM in final binary to reduce as possible use of emulation by coldfire that  slow down the machine. Since people with good knowledge have done the job for coldfire. But my point of view less there is ASM better it is, there is no interest for such code to do it in ASM, if compiler is good enough difference should be very small (ok GCC 4 and higher not give good code for such code but this is not a reason), my point of view better is to remove ASM code ;-), with new processor optimization can change (we see this on 68080 for example for some of my code on 68060 GCC 4 give better result than GCC 3 but on 68080 GCC 3 is faster than GCC 4)

mikrosk:
> my point of view less there is ASM better it is, there is no interest for such code to do it in ASM, if compiler is good enough difference should be very small

Normally I would 100% agree with you but here I'd like to go forward and remove the C code instead:
- it's just an isolated change, i.e. `C_ONLY` doesn't replace all asm routines, just one
- for 000 and 030 targets it can be still slower than the asm version (there was a reason why `quickzero` and friends had been added in the first place)
- I doubt there is someone willing to do a rich set of performance tests to make sure that we wont make the kernel slower by this change

th-otto:
quickzero is a highly optimized routine, and (amongst others) used to clear the BSS and/or TPA upon program exec. Replacing this with a generic C version would certainly be noticable, and since it is already adapted to coldfire, i don't see a reason to replace it.

Note that even on modern CPUs, memset() etc. are typically implemented in ASM too in the C library. This is just a time-critical function, especially on large memory blocks.

PR #271, opened 2022-07-23
Cache converted textures

### LINKED ISSUE #180: XaAES could cache converted textures
This is clearly a feature request. While testing #171 I have noticed how awfully inefficient the texture handling is - it tries to load & convert like 20 images, one by one, so even on 32 MHz it takes good 10 seconds (!) to load.

XaAES could easily store prerendered textures into its `width` `height`.`depth` folder which has already some use.

mikrosk:
Contributed by Xavier Delatour.

Fixes #180.

### DIFF
--- a/doc/xaaes/xaaes.u
+++ b/doc/xaaes/xaaes.u
@@ -178,7 +178,9 @@ XaAES also reads IMG files from the folder /xaaes/img/. This is needed if you
 want the textured look. If not present, you get the default rendering.
 This does not affect the use of the gradient setting in xaaes.cnf.
 If you use gradients or don't want to use the textures at all, set option
-"textures = 0" in xaaes.cnf, so XaAES will boot faster.
+"textures = 0" in xaaes.cnf, so XaAES will boot faster. However using the
+cache enabled with "textures_cache = yes" makes the speed difference
+neglible.
 
 For more information on texture files, see Theming and Extended Textures.
 !end_node # IMG Files
@@ -408,18 +410,23 @@ When true, standard window widget are automatically highlighted when the
    actually highlighted, they are just animated, that is, redrawn in SELECTED
    state. E.g. widget_auto_highlight = true
 
-!item [GRADIENTS <number> (default 0)] (!nl)
+!item [GRADIENTS <name|[0|1]> (default 0)] (!nl)
 Setting gradients to (!B)default(!b)/(!B)1(!b), or the name of a file in the gradients
    folder (/XAAES/GRADIENT/*.GRD). This allows the use of colour gradients
    in the window widgets/AES object renderer in hicolour or truecolour video
    modes. This replaces the use of texture images. E.g. gradients = n_yellow
 
-!item [TEXTURES = <name|[0|1]> (default = img)] (!nl)
+!item [TEXTURES <name|[0|1]> (default 0)] (!nl)
 Setting textures to a nonzero value or a name turns on the use of textures
    in the window widget and aes object renderer in color modes.
    This is the name of a texture-folder in /xaaes/ directory.
    If you use gradients, set this to 0, XaAES boots faster then.
    E.g. textures = img or textures = 0
+
+!item [TEXTURES_CACHE <bool> (default false)] (!nl)
+Cache textures for current bit depth in <aes_home>/cache. If you update
+   the texture files or change the palette, delete the cache folder, too.
+   E.g. textures_cache = yes
 
 !item [PALETTE <character-sequence> (default (!I)internal(!i))] (!nl)
 Load a specific palette from /XAAES/PAL/*.PAL, resulting in a tra!-di!-ti!-onal
@@ -429,7 +436,7 @@ Load a specific palette from /XAAES/PAL/*.PAL, resulting in a tra!-di!-ti!-onal
 Change the default background colour, colour numbers are in standard colour
    palette order. E.g. back_col = 13
 
-!item [MENUBAR <bool> (default 2)] (!nl)
+!item [MENUBAR <number> (default 2)] (!nl)
 Hideable menubar, where "0" is off, "1" is on, and "2" is always on (default).
    E.g. menubar = 0
 
--- a/xaaes/src.km/cnf_xaaes.c
+++ b/xaaes/src.km/cnf_xaaes.c
@@ -164,6 +164,7 @@ static struct parser_item parser_tab[] =
 	{ "GRADIENTS",		PI_R_T, cfg.gradients,       { dat: sizeof(cfg.gradients) } },
 #endif
 	{ "TEXTURES", 		PI_R_T, cfg.textures,       { dat: sizeof(cfg.textures) } },
+	{ "TEXTURES_CACHE",	PI_R_B, &cfg.textures_cache		},
 	/* startup things */
 	{ "SHELL",                 PI_V_ATK, pCB_shell			},
 	{ "RUN",                   PI_C_TA,  pCB_run			},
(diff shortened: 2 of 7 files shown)


mikrosk:
@xdelatour thank you for looking into this! I can confirm that the speedup is pretty noticeable. From ~40s (~10s takes the kernel & modules) loading time went down to ~28s, so basically loading time got cut by one third! And it seems it could be even more, see below.

Also, saving the bitmaps is barely noticeable, the first run takes about 1-2 seconds longer against clean `xaaes.km`.

I tested it on my TT@32 MHz with pretty quick TT-RAM & IDE (the Thunder/Storm combo). I have a few comments on this:

- The need for long names support seems to be unnecessary. We already have a folder `<xaaes>/widthheight.depth` which stores the background image for the desktop. This virtually begs for usage for your cache system.
- I think the usage of xattr's dev/index is also unnecessary, you can easily use just a filename. I.e. your cache entry would look like `<xaaes>/12801024.16/slider.img`
- No need to worry about the 8b/hc subfolders or different palettes; it's totally OK to ask the user to delete the cache if he changes the palette in xaaes.cnf (or, if you want to be really clever, you can put the palette _itself_ into to cache folder and compare whether the cache is still valid)
- There seems to be some discrepancy when handling some of the cache entries. Take a look at [xa_boot.log](https://github.com/freemint/freemint/files/9173456/xa_boot.log) (logged the first and second run): files `hc\slider.img` (`000200001358.16`), `hc\slide.img` (`00020000135E.16`), `hc\info.img` (`000200001356.16`), `hc\wtitle.img` (`000200001357.16`), `hc\exterior.img` (`000200001360.16`) are correctly not found at first but at the second attempt they are reported as invalid which seems wrong as they were created just seconds ago.

mikrosk:
For clarification, this is my changed `image_cache_get` function, to see where the debug outputs come from:

```
static void
image_cache_get(char *name, XATTR *xattr_now, XAMFDB *mimg)
{
	struct file *f;

	mimg->mfdb.fd_addr = NULL;
	f = kernel_open(name, O_RDONLY, NULL, NULL);
	if (f)
	{
		long bmsize, bmread;
		int invalid;
		XATTR xattr_src;

		kernel_read(f, &xattr_src, sizeof(XATTR));
		xattr_now->atime = 0x5843; xattr_now->adate = 0x3031; /* Magic value: 'XC01'*/
		invalid = memcmp(&xattr_src, xattr_now, sizeof(XATTR));

		if ( !invalid )
		{
			BLOG((0,"\r\nAn effort to alloc %s", name));
			kernel_read(f, mimg, sizeof(XAMFDB));
			bmsize = (long)((long)mimg->mfdb.fd_wdwidth * (mimg->mfdb.fd_nplanes == 15 ? 16 : mimg->mfdb.fd_nplanes) * mimg->mfdb.fd_h) << 1;
			if ((mimg->mfdb.fd_addr = kmalloc(bmsize)))
			{
				bmread = kernel_read(f, mimg->mfdb.fd_addr, bmsize);
				if (bmread != bmsize)
				{
					kfree(mimg->mfdb.fd_addr);
					mimg->mfdb.fd_addr = NULL;
					invalid = 1;
				}
			}
		}
		kernel_close(f);
		if ( invalid )
		{
			BLOG((0,"\r\nInvalid file %s", name));
			_f_delete(name);
		}
	}
	else
	{
		BLOG((0,"\r\nFile %s not found", name));
	}
}
```

xdelatour:
Thanks for your feedback 
I'm very busy these days but I'm still working on it.

xdelatour:
* `<xaaes>/widthheight.depth` vs `/var/cache/xaaes`: if sometimes we are using 640x480x16 and later 80x600x16, the same file will be stored twice
* `<dev><index>` vs `filename` : if the user changes from one theme to another, we can't know if `filename` comes from theme1 or theme2
* If we assume Xaaes only loads IMG from theme (this is the case for now) and theses files are immutables, and the user is aware that he must clear the cache when changes are made (palette and/or textures settings in xaaes.cnf), then I think I could use something like `/var/cache/xaaes/<filename>.depth`
* The cache may be invalid if:
  * the attributes (of the IMG file to load) do not match those stored in the cache file. it means the source file has changed
  * It is not possible to read the correct number of bytes

I added log messages to show attribute and size differences if there are any

[xaaes02060.km.zip](https://github.com/freemint/freemint/files/9175847/xaaes02060.km.zip)

skaftetryne:
I never use textures so I can't say anything about the usability of such a cache. However, I think it's very unfortunate that a kernel module feature like this depends on the presence of a folder that is not guaranteed to be available. u:\var is only available when it has been explicitly set up by the user. This cache should be located in the same folder as the XaAES kernel module. Also, I have not checked if this is already the case but please make sure that all files fits in the 8+3 format.

xdelatour:
Here is a new version:

* [xaaes02060.zip](https://github.com/freemint/freemint/files/9196816/xaaes02060.zip)
* [textures_cache-v2-20220726.patch.txt](https://github.com/freemint/freemint/files/9196825/textures_cache-v2-20220726.patch.txt)

The diff is based on master, not your branch (I'm definitely not familar with git)

Changes:
* long names: unneeded, now are `<filename>.<depth>`
* files are stored in `<aes_home>/cache` (automatically created)
* There is no way to invalidate files if theme or palette is changed (XATTR is no more used and all themes use the same filenames, it's not possible to check if the file matches the source .img)

mikrosk:
Today I could finally test it, works perfectly. Some numbers:
- current master: 38s
- first run of your latest version: 39s
- second run of your latest version: 20s
- textures off: 19s

As you can see, this patch makes a hell of a difference, at least on my TT@32 MHz. Basically makes the texture loading invisible.

I've cleaned it up a bit, added documentation (and cleaned up it a bit, too) as well as fixed the default state for `textures` (if was compared against `'0'` but default state is binary zero).

Please take a look in the changes tab and if you don't see anything wrong, I can merge it.

shoggoth77:
Wow! I welcome this feature, and I've often contemplated about it myself.

One small thing I've thought about (that may be a "feature" for now) is that some machines sport multiple pixels formats for the same screen depth. E.g. 8bpp chunky vs 8bpp planar, which exist in the wild today (Supervidel). I think the same issue may arrise for the V4SA (which could potentially have 3 different 8bpp formats, and also multiple 16/15-bit ones).

No biggie in practice, but it's something to think about. All in all - hats off and thanks the effort!

mikrosk:
@shoggoth77 isn't this actually solved automatically? I don't know how the conversion functions work exactly but I'd imagine if it is native format which they output in the end, it would be stored like that as well.

So sure, if you cache chunky 8bpp and set bitplane 8bpp resolution you have to clean the cache but otherwise... it should work out of the box.

shoggoth77:
@mikrosk this is exactly the (very) special case I was referring to.

xdelatour:
XaAES saves the wallpapers in a folder whose name only contains the color depth. (640480.16 for example)
If the image is loaded in a 16-bit mode with a different pixel format, I imagine there will also be display issues.
On which machine or emulator can this be tested?

Note: the bitmap loading function takes into account the pixel format but only for modes higher than 8 bits

In the patch, change:
`sprintf(strrchr(cache_path, '.')+1, 4, "%d", screen.planes);`
with:
`sprintf(strrchr(cache_path, '.')+1, 4, "%c%d", screen.pixel_fmt+'c', screen.planes);`
This will add a character (a..k) according to the pixel format (-2..8 as returned by [detect_pixel_format()](https://github.com/freemint/freemint/blob/66b4e55eb6a0366ab82c6e8f8f00b0f2e630f354/xaaes/src.km/trnfm.c#L2368))

mikrosk:
If there is ever need for more advanced functionality, we can add it. For now, I'm happy with how it works as is.

@xdelatour many thanks for your contribution!

PR #274, opened 2022-08-05
Print error message instead of crashing

mikrosk:
For the hundredth time I spent half an hour of my life thinking why svethlan.xif is crashing on my CT60. And it is because it calls `ct60_vmalloc` provided by `SV_XBIOS.PRG`. Without `SV_XBIOS.PRG` present it crashes on a fatal error without any further information.


th-otto:
Shouldn't this be merged?

But i would prefer to use \r\n in the messages. \n\r is imho nonsense and not a valid line ending.

mikrosk:
Indeed, I have been postponing this for a final test, done now. Also changed the line endings and did a couple of improvements.

Issue #273, opened 2022-07-30
asix.xif leads to random crashes/freezes


mikrosk:
I started noticing this when I was putting more stuff info mint.cnf, namely:
- usb `loader.prg`
- gluestik.prg
- `ifconfig` & `route` commands

When I put those three lines for the first time, it would crash all the time. Then I (out of despair) tried to replace the `ifconfig` and `route` commands with `/bin/sh /bin/network.sh` (where `network.sh` would contain basically the same thing) and crashes stopped.

Then I tried to compile a 030 version of asix.xif and I got the crashes back again.

Sometimes booting would succeed but when xaloader is starting xaaes, it crashes on a null pointer.

And sometimes it's really random... for instance, this is when `gluestik.prg` is about to be executed:

![PXL_20220730_222930575](https://user-images.githubusercontent.com/3062260/182002422-a3c1b578-ec78-40ce-a765-aa451d7ff545.jpg)

(the crash/assert is always the same)

Another interesting thing is that _if_ it succeeds, then the setup is rock solid. It really looks like some kind of race condition during init. But not module init itself -- I can for instance remove all three lines from mint.cnf, start `loader.prg`, `gluestik.prg` and `network.sh` from deskop and it would sometimes crash even there. So it must be somehow related to assigning the network interface or something like that.

mikrosk:
Hmmm... looking into asix.c has revealed some interesting information:
- the assert comes from kernel's `kfree` function and `asixpoll` does indeed calls kfree() in `asix_poll_thread` (which is in turn awakened from `asix_poll`)
- `asix_poll` is initialised from `asix_recv` to be called every 50ms on 68000 and every 10ms on 68020+ (i.e. the more frequent crashes with my 030 build make sense!)

czietz:
That is a very interesting observation. Yesterday, while working on my
(not yet published) driver based on asix.c, I was seeing the exact same
thing: Random crashes, sometimes when the network was brought up by
ifconfig and route, sometimes right after it during xaloader. Same as
with you: _If_ it boots successfully, it runs totally solid.

I thought this had to be a bug in my driver and unsuccessfully reviewed
my changes again and again. But since you see the crashes with the asix
driver, too, this gives me further ideas.

czietz:
Okay, my current theory: What if `asix_recv()` is called via the `asix_poll_thread` while the network stack is still in the process of bringing the interface up? As you say: a race condition.

What seems to fix the problem for me, is thus:
[...]

PR #275, opened 2022-08-06
Disable strict aliasing

mikrosk:
When enabled (what is with -O2 by default), one can observe unexpected and random freezes when downloading and uploading larger chunks to/from Atari (typically it is visible with files >10 MB).

This has been reported numerous times and basically all reports claim that this wasn't the case with older kernels. And indeed, 1.16.1b kernel doesn't suffer from this issue while 1.17.0 does.

So what changed? The official compiler did. 1.17.0 and all later releases have been compiled with 4.6.4 and later 7.5.0. And since it was gcc 4.x which introduced the aggressive optimizations... we have the answer.

Change the optimization level for the whole project, later we can decide what parts can have it enabled again.

### DIFF
--- a/CONFIGVARS
+++ b/CONFIGVARS
@@ -86,7 +86,7 @@ SED   = sed
 #
 # optimizations
 # 
-OPTS = -O2 -fomit-frame-pointer
+OPTS = -O2 -fomit-frame-pointer -fno-strict-aliasing
 
 #
 # warning options


vinriviere:
This is very sad. But better than an unreliable kernel.
Strict aliasing indeed allow more aggressive optimizations. That would be a shame to not been able to get benefit from that. I hope that real cause of trouble will be found.

mikrosk:
To give you some context... even if it wasn't me who found the issue in #273, thanks to that investigation I have noticed how many "dirty" casts we are doing in our networking code (actually it was one of the ideas I had, to check the USB part with `-fno-strict-aliasing` but it didn't help).

So now that we have fixed USB stack, I wanted to compare (and basically check whether they still work...) some of the networking gear I have - NetUSBee, EtherNAT, Svethlana, to see how big impact the USB middleware has.

And what have I found? The random freezes (soft = one can terminate `ftp`, `litchi` etc and hard = whole system frozen) reported so many times, most recently here: https://www.atari-forum.com/viewtopic.php?p=401584#p401584

So I was like OK, I wasted so much time on benchmarking and looking into code, what about verifying this claim? So I downloaded 1.16.1b, 1.17.0 and 1.18.0 releases, put the same `ifconfig` and `route` into mint.cnf, copied the same (latest) hardware drivers (svethlana.xif, ethernat.xif, ne2kect2.xif) and all latest tools (`ifconfig`, desktop, toswin2, ...) into SYSDIRs and... very early I found out that 1.16 is the only package working reliably. And since the hardware drivers were the same, it must be the mint/mintnet itself.

I had a hunch that it couldn't be something changed so drastically between 1.16 and 1.17 so I decided to try a bold move -- compile latest kernel with gcc2. It wasn't that hard in the end most importantly... replacing `mint.prg` and `inet4.xdd` immediately led to rock solid experience!

I had three test cases:
- `ftp` + `get` 85 MB file from a Linux PC
- `ftp` + `put` the same file to the same Linux PC
- `wget -O /dev/null` as seen here: https://www.atari-forum.com/viewtopic.php?p=294172#p294172

Regardless of the hardware, I would **never** pass all three of them. Sooner or later the transfer pauses endlessly / freezes / times out.

With gcc2 and later gcc7 + `-fno-strict-aliasing` I can do as many `get`s and `put`s as I like, impossible to achieve before.

It seems that the only one app is equally unreliable as before: Litchi. While download to Falcon works perfectly (it didn't), upload sometimes hangs the whole machine. I'll do more tests later.

mikrosk:
@czietz since you've been working on network drivers, what kind of tests you do? You have never noticed such mysterious freezes with your hardware? (it really requires several tens of MBs to notice, usually the first 20-30 MB works).

mikrosk:
Oh and final remark: this is not so crazy / bad change as it might seem. As far as I could see, all major kernels are compiled with `-fno-strict-aliasing`, Linux [included](https://github.com/torvalds/linux/blob/master/Makefile#L524).

czietz:
I never consciously noticed this. But then, my usual test file is about 4 MB. I now tried wget with that 100 MB file, and guess what: the download froze after about 25 MB.

If you can reduce the issue to one single module (e.g., inet4.xdd), I'll use my Ghidra and BinDiff skills to find out where gcc does actually produce different code with and without -fno-strict-aliasing.

mikrosk:
@czietz so again some bitter-sweet victory. ;)

Is knowledge that it is mint.prg & inet4.xdd enough? Because when only one of them is replaced, the situation gets better but not _perfect_. For instance, with only inet4.xdd gcc2 compiled, I was able to download whatever I wanted but uploads were still broken.

czietz:
😢  - I was hoping you would say something like "inet4.xdd is enough". BinDiffing the whole MiNT kernel will be quite some work. I mean: not the diffing itself, but checking for each function whether the change is relevant to our problem.

mikrosk:
Maybe there's a reasonable alternative. We know for sure that inet4.xdd is broken, right? So, if you fix that, we will see what kind of code got fixed and why. And we can apply that knowledge to mint.prg and maybe even usb stuff and see what happens.

ps68060:
I've never seen this freezing.

I sometimes transfer photos from my phone to the Milan via smb. There might be 10+ jpg files, so 50-70 MB at a time. Does it have to be one large file ?

vinriviere:
BTW: IIRC, strict aliasing issues were (partly) fixed in the FreeMiNT kernel by using the "union" keyword.

In EmuTOS, we used a different approach. We didn't modify the code much, but added `__attribute__ ((__may_alias__))` to problematic places. With a few smart macros, that's pretty straightforward.
https://github.com/emutos/emutos/blob/master/include/portab.h#L194

Of course, the biggest difficulty is to find the problematic issues. With EmuTOS this was rather simple, as it failed unconditionally. I'm a bit surprised that issues are so much random with FreeMiNT. I'm not sure that strict aliasing is the real source of the issues. They are probably coupled to some uninitialized memory, like the recent issue #273.

czietz:
Argh, now I can't reproduce the failure anymore. Without changing anything on my TT, I had multiple 100 MB downloads complete successfully yesterday evening. 😢 

Anyway: I started BinDiffing inet4.xdd and almost right away I found something scary (although it probably not the cause of the hanging downloads). Left side is with `-fno-strict-aliasing`, right side is the default without this option:

grafik

This is along these lines of code:
https://github.com/freemint/freemint/blob/a54c65e5fda06e40f73b91a4cfcfd57c2d181bbb/sys/sockets/inet4/tcputil.c#L261-L265

Notice the missing `CLR.W` on the right side? Without `-fno-strict-aliasing`, gcc optimized `otcph->chksum = 0;` away! This is blatantly wrong, because the TCP checksum calculation relies on that field being 0 to generate the correct checksum.

Now you might ask: Why did this ever work? `tcp_sndack()` and the similarly affected `tcp_sndrst()` are only sparsely used. In particular, the normal ACK generation during a TCP transmission is handled elsewhere.

Finally to why assumed strict aliasing rules cause gcc to do this "optimization": `tcp_checksum()` is written in inline assembly and the constraints are such that gcc assumes that the inline assembly only works on the pointer to the TCP header (`otcph`) but does not dereference it. Since by strict aliasing rules nothing else can point to the same data, gcc can (wrongly) assume that the memory pointed to by `otcph`, including `otcph->chksum = 0`, is never read and therefore the assignment can be "optimized" away.

This can be perhaps be exemplified with Compiler Explorer: https://tinyurl.com/23ybrp8j. Notice, how there in `test1()` the assignment of the value 42 is incorrectly removed. With proper inline assembly constrains in `test2()` it isn't. In other words, the inline assembly constraints were always wrong, but turning on strict aliasing rules triggered this bug. 

Considering how many of such subtle bugs might lurk in the codebase, I'm all for accepting this PR and disabling strict aliasing rules.

vinriviere:
Good catch again, @czietz 😃

I wasn't aware of such usage of the "m" constraint to pass whole structures. Pretty weird. And I'm not sure about correct assembler syntax to access the struct members at an offset greater than 0.

However, the classic way to solve this issue is to add "memory" to the extend asm clobber list.
Classic correct implementation:
```
static int classic_correct_implementation(struct foo *bar)
{
    int ret;
    __asm__ (
        "move.w (%1),%0"
        : "=d"(ret)
        : "a" (bar)
        : "memory"
    );
    return ret;
}
```

The GCC documentation about [Clobbers and Scratch Registers](https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers) is explicit:

> The "memory" clobber tells the compiler that the assembly code performs memory reads or writes to items other than those listed in the input and output operands (for example, accessing the memory pointed to by one of the input parameters). To ensure memory contains correct values, GCC may need to flush specific register values to memory before executing the asm. Further, the compiler does not assume that any values read from memory before an asm remain unchanged after that asm; it reloads them as needed. Using the "memory" clobber effectively forms a read/write memory barrier for the compiler.

Note that FreeMiNT sources (and EmuTOS as well) use the AND_MEMORY macro to add the "memory" clobber, in order to deal with different compiler versions.

vinriviere:
> Notice the missing `CLR.W` on the right side? Without `-fno-strict-aliasing`, gcc optimized `otcph->chksum = 0;` away! This is blatantly wrong, because the TCP checksum calculation relies on that field being 0 to generate the correct checksum.

I already said it above, but excellent find @czietz  😃

BTW: I'm not surprised about this checksum issue. I occasionally use Wireshark to debug network issues with ARAnyM or FireBee. Many times, I saw Wireshark warnings saying "bad checksum". I always wondered if they were real problems or false alerts. Generally, that didn't hurt, and the transfers still succeeded. Maybe just because of TCP retries. This could be checked again before and after the fix proposed here. As you said, the real problem probably triggers only in special situations.

Note that I feared a mistake when I patched the assembler checksum routine for ColdFire. But after inspection, I went to the conclusion that the ColdFire checksum routine was correct. The network traces were similar between the FireBee and ARAnyM.

czietz:
Another function that changes quite a lot is `tcp_dropsegs()` in https://github.com/freemint/freemint/blob/master/sys/sockets/inet4/tcp.c. At first glance, I don't understand the exact nature of the changes that gcc introduces. (And `m68k-atari-mint-objdump -S` struggling with this object does not help.)

However, a bug in this function might explain why downloads hang. It takes care about freeing receive buffers that have been handled and about telling the sender that we can receive new data.

@mikrosk : Maybe you can try just compiling tcp.c with `-fno-strict-aliasing`. Since I cannot reliably reproduce the bug I can't test myself.

czietz:
> Many times, I saw Wireshark warnings saying "bad checksum".

But be sure to check the Atari side. You might see spurious "bad checksum" warning for packets sent by the PC if the checksum calculation is offloaded to the network adapter hardware.

mikrosk:
> Argh, now I can't reproduce the failure anymore. Without changing anything on my TT, I had multiple 100 MB downloads complete successfully yesterday evening. 😢

The irony is that the same applies to my testing - NetUSBee basically worked on all kernels, even Litchi miraculously works where it didn't before. Even Ethernat and Svethlana were 90% of cases working reliably. So your report of having seen the same problem on your setup was the only thing keeping me apart from suspecting that this is completely unrelated to strict aliasing or that maybe it applies to high throughput network cards (both Ethernat and Svethlana go above 1 MB/s).

So certainly there is some element of randomness involved, as Vincent mentioned.

Anyway thanks for looking into it, your findings are indeed scary! I'll try to do a few more tests to be able to confirm/dismiss possible fixes.

th-otto:
As Vincent said, the "wrong" optimization should be avoidable by adding a 
"AND_MEMORY" clobber to the list. I think that is already the case for most 
inline assembly. However, unless compiling with -fstrict-aliasing gives 
important optimizations, i'm in favour for just adding -fno-strict-aliasing to 
the global CFLAGS.

As other have noticed, this is already the case for other kernels, including 
linux. I guess for similar reasons.

That might not solve all problems, but certainly explains observations that 
newer kernels have become more unreliable over the time.

Another thing  that i would prefer is to move back to build the snapshots with 
gcc-4.6.4. Except for better warnings (which we can always get by compiling it 
locally with different versions of gcc), i don't see any advantage of using 
gcc 7.5.

mikrosk:
> I've never seen this freezing.
> 
> I sometimes transfer photos from my phone to the Milan via smb. There might be 10+ jpg files, so 50-70 MB at a time. Does it have to be one large file ?

Perhaps your use case really doesn't apply here: it should be ideally one huge file, using direct TCP/IP connection (ftp, scp, ...) and not on Milan. :) As @skaftetryne reported [here](https://www.atari-forum.com/viewtopic.php?p=401603#p401603), [here](https://www.atari-forum.com/viewtopic.php?p=401647#p401647) and [here](https://www.atari-forum.com/viewtopic.php?p=401749#p401749), it would seem that you either can have a Milan and happily use whatever you want or have a Falcon and happily use nfs (perhaps samba, too). But ftp + falcon 060 + ethernec didn't work.

czietz:
I start to wonder whether my initial observation (in https://github.com/freemint/freemint/pull/275#issuecomment-1207255507) was some sort of glitch; perhaps with the server on the other side. Over the past days, whenever I was testing and refining my network adapter, I also tested large file downloads: no more freezing downloads.

Regardless of that and as proven by my findings above, "strict aliasing" optimization triggers bugs in the code and is perhaps best switched off.

mikrosk:
As I was observing suspicious freezes on my Falcon, even more suspiciously only appearing when launching `ifconfig` and/or `route` commands from mint.cnf, I'd say let's give this one a try.

Most likely this wont solve _all_ networking / freezing issues but at least everybody will have the same base and we can compare results.

mikrosk:
Adding for visibility: @skaftetryne confirmed that the new builds (with this change applied) indeed improved his network experience: https://www.atari-forum.com/viewtopic.php?p=454314#p454314.

PR #276, opened 2022-09-01
PicoWifi support (USB-to-Wifi adapter)

czietz:
I just released the PicoWifi (https://github.com/czietz/picowifi/), a GPLed USB-to-Wifi adapter, for which I wrote this FreeMiNT driver. Since it is outside of the Lightning/USB source tree, I did not want to push it without previous review; for which I'm creating this PR.


th-otto:
SRCFILES seems to have only picowifi.c added, but i see the whole file in the diff. Could it be that there is some change in linefeeds?

czietz:
Good find! Fixed this now.

PR #280, opened 2022-09-29
V4SA Ethernet driver

agranlund:
FreeMINT ethernet driver for Vampire V4SA.


th-otto:
Many thanks for that.

Of course i cannot test it, but it compiles, so i'm just merging it.

Just 2 things i noticed while taking a short look:
- The value after the magic "XBRA" is not a cookie, but just an ID who is using that vector. You should use a value other than 0 for that.
- line 73 of v4net_int.S should imho rather read ```move.l _v4net_int_old,-(sp)```

PR #282, opened 2022-10-10
Remove proc::in_kern flag

mikrosk:
While investigating a fix for https://github.com/freemint/freemint/commit/b3214578 I noticed one inefficiency - we call [enter_kernel()](https://github.com/freemint/freemint/blob/master/sys/arch/kernel.S#L63) basically for no reason. Instead of just `bset #0x7,_in_kernel` we additionally set the same flag for the current process. However this process flag is never really used, except one debug output.

Considering how often this function gets called from trap handlers it could save quite a few cycles.

### DIFF
--- a/sys/arch/kernel.S
+++ b/sys/arch/kernel.S
@@ -17,7 +17,6 @@
 #include "magic/magic.i"
 
 	.globl	_enter_gemdos
-	.globl	_enter_kernel
 	.globl	_leave_kernel
 
 	.extern	_curproc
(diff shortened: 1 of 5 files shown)


mikrosk:
Please take a closer look. That's a global variable called `_in_kernel`, not per-process variable `in_kern`.

However good point with the struct size, I should have put `short unused;` there - as it was [before](https://github.com/freemint/freemint/commit/bab5567f99082bf8ee0d823b7a7255874989727c#diff-9b4a111356e7d1677570f45bfad28099aad757f85c29e20195b02e0dbb29ad00L254) I introduced `in_vdi`.

PR #284, opened 2022-10-13
Restore d0.l in trap2 handler

### LINKED ISSUE #283: CFLib-compiled apps crash with memory protection if NVDI/fVDI is not present
No, this is not a regression from fixing #22 :) To my surprise, it has been here all along - before my fix and also after my fix. How so? Well, it's reproducible only on 040+ machines. I personally tested it on Aranym and CT60, same symptoms.

Fortunately, repro is easy:
- download latest freemint snapshot for aranym
- disable `fvdi.prg` in `mint.cnf` and set `video = 26` in `xaaes.cnf`
- boot  with aranym-mmu

Desktop loads fine but TosWin2 is nowhere to be found. Double-click on toswin2.app, nothing. Double-click on qed.app, nothing.

When compiled with debug info:
```
pid   6 (toswin2): kern_procdir_get_stat: enter
pid   6 (toswin2): kern_procdir_get_stat: do ksprintf
pid   6 (toswin2): get_cookie(): tag=4D694E54 (MiNT) ret=1290DFC
pid   6 (toswin2): get_cookie(): value 00000113 returned at 1290DFC
pid   6 (toswin2): GET_COOKIE($4D694E54): r=0, val=$113
pid   6 (toswin2): get_cookie(): tag=4D616758 (MagX) ret=1290DFC
pid   6 (toswin2): get_cookie(): lookup failed
pid   6 (toswin2): GET_COOKIE($4D616758): r=-1, val=$FFFFFFD6
pid   6 (toswin2): get_cookie(): tag=4D675043 (MgPC) ret=1290E08
pid   6 (toswin2): get_cookie(): lookup failed
pid   6 (toswin2): GET_COOKIE($4D675043): r=-1, val=$FFFFFFD6
pid   6 (toswin2): get_cookie(): tag=6E414553 (nAES) ret=1290E00
pid   6 (toswin2): get_cookie(): value 0112A000 returned at 1290E00
pid   6 (toswin2): GET_COOKIE($6E414553): r=0, val=$112A000
pid   6 (toswin2): get_cookie(): tag=4E564449 (NVDI) ret=1290E08
pid   6 (toswin2): get_cookie(): lookup failed
pid   6 (toswin2): GET_COOKIE($4E564449): r=-1, val=$FFFFFFD6
pid   6 (toswin2): Mxalloc(1204,30)
pid   6 (toswin2): attach_region: reg 4, return loc 42000 (42000)
pid   6 (toswin2): [1][ PROCESS  "toswin2"  KILLED: | MEMORY VIOLATION.  (PID 006) | Type: hardware  PC: 00E57A64 | Addr: 6104487A  BP: 01246000 | 
pid   6 (toswin2): signal #10 raised [syscall_pc 0x12655E0, exception_pc 0xE57A64]
pid   6 (toswin2): ufree(0x11B3F4C, XaAES:../xa_appl.c,862)
pid   6 (toswin2): ufree(0x11B3F30, XaAES:../xa_appl.c,864)
```

This crash and PC is repeated for every attempt. First I thought whether that 0x30 mode (supervisor protected) doesn't do some damage there but nope, there are other allocations with this mode. And then it hit me - what if it is the startup-code? And indeed, if one starts COPS or HypView from the snapshot, they work fine. Only QED and TosWin2 crash.

From EmuTOS map file I could see that it is crashing somewhere in `_vdi_vst_load_fonts` and indeed, it really crashes at this place: https://github.com/freemint/cflib/blob/master/appinit.c#L94

So ...

- why does it crash only on 040+ machines?
- is the check for GDOS presence above it not sufficient?
- what's so special about this function that it crashes in both EmuTOS and CT60 TOS (4.x) ?
- why does it crash at all?

mikrosk:
While VDI opcodes are 16-bit, there is one special case: -2 (`vq_gdos()` / `vq_vgdos()`). While `vq_gdos()` manually extends the (word) return value in d0 to long, `vq_vgdos()` doesn't touch it after returning from the OS.

`vq_vgdos()` is supposed to return a 32-bit GDOS identifier but it is the *word* -2 (passed as an input parameter) which the programmer is supposed to check against to see if a GDOS is installed.

Many, if not all, `vq_vgdos()` bindings (gemlib's in particular) just return d0.l and that's the issue here -- if `vq_vgdos()` returns a 32-bit int and documentation says "-2 means GDOS not installed", then there's no reason to expect that the -2 should be first down-casted to a (16-bit) short by the programmer.

Since TOS and EmuTOS doesn't touch d0.l *at all* if -2 is passed to trap2, FreeMiNT should do the same.

Fixes #283.

### DIFF
--- a/sys/arch/syscall.S
+++ b/sys/arch/syscall.S
@@ -743,7 +743,7 @@ norm_trap2:
 	// A1 set to callers stack by build_context
 	move.l	_curproc,a0
 	move.l	(a0),sp			// fetch our systemstack
-	move.w	6(a0),d0		// D0.w saved by build_context, function code (AES or VDI)
+	move.l	4(a0),d0		// D0.l saved by build_context, function code (AES or VDI)
 	move.l	8(a0),d1		// AES/VDI pb
 	cmp.w	#115,d0			// VDI?
 	beq	enter_vdi


th-otto:
Sorry, but that is wrong. d0.w is the *input* opcode to the VDI call, and that is a word, also for vq_vgdos call. Your patch imho won't fix anything.

mikrosk:
Thorsten, WTF. Do you really think I did not test it? Of course it fixes the issue.

gemlib passes -2 in d0.l and it expects the same value on the return from OS so it can return back to the callee.

th-otto:
Well maybe it does fix the symptom, but the input opcode to vdi is a just a word.

The way qed and/or toswin do the check is broken, and they really should use vq_gdos() instead vq_vgdos() (with some explanations maybe, and a link to the issue). It's true, TOS just returns d0 unmodified. It doesn't even do an extra check for the -2 case. But that also means, it just returns whatever value was there in the upper 16 bits.

You could also argue that the binding of vq_vgdos() in gemlib is wrong. It should sign-extend the result, if -2 is returned. But your patch just forces everyone to use -2.l as opcode for the vq_vgdos call, and that is wrong.

th-otto:
BTW, syscall.S should really use some constants defined in magic.i, eg. P_SYSCTXT

mikrosk:
> Well maybe it does fix the symptom, but the input opcode to vdi is a just a word.

That symptom is the whole problem here. :-) There is non-documented behaviour that TOS doesn't touch D0.l if D0.w == -2. That is what this patch aims to do and it is its sole purpose. 

> But your patch just forces everyone to use -2.l as opcode for the vq_vgdos call, and that is wrong.

... and it is also not true. :) `move.w #-2, d0`, `trap #2`, `cmp.w #2,d0` will work just as good.  Where do you see something else?

I absolutely agree with you that it's much safer to use `vq_gdos()` instead of relying on such behaviour (or quirks introduced by different libraries) but as you say, it has to be worked around at least to the TOS compatibility level:

>  to support old bindings that expect the upper bits to be preserved

Right now, every TosWin2 and QED build in existance contains this bug. And there surely is at least a few other misguided apps/libraries which do the same (i.e. `moveq #2,d0` & `tst.l d0`)

> the binding in vq_vgdos in gemlib should be fixed, too.

Agreed, it works both ways, someone can get his hands on newer QED/TosWin2 and still use an older kernel.

PR #289, opened 2023-01-10
Use latest gemlib's constants

mikrosk:
To be merged when @vinriviere updates his toolchain with latest gemlib to avoid compilation errors.


th-otto:
Is that really needed? It only removes some ifdefs, but even if vincent updates his toolchain, that does not mean that users immediately install his newer version.

mikrosk:
I wouldn't say needed but it's always nice to remove code. :-) And if Vincent updates his toolchain, he will also update his PPA, therefore the fix is easy: apt-get update && apt-get upgrade (although I'm not sure about cygwin, is it possible to update its mint packages as well?)

vinriviere:
OK, I will update my Ubuntu and Cygwin packages with new MiNTLib.

vinriviere:
OK, done. I've updated my Cygwin and Ubuntu packages with new MiNTLib:
http://vincent.riviere.free.fr/soft/m68k-atari-mint/
http://vincent.riviere.free.fr/soft/m68k-atari-mint/ubuntu.php

EDIT: Damn, that issue was about **GemLib** 😭
Well, I will have to do another round.

xdelatour:
Don't worry , there are also a recent fix in mintlib (opkg package manager  compiles fine but doesn't work because of two bugs including basename which was fixed 5 months ago)

vinriviere:
Done again. I've updated my Cygwin and Ubuntu packages with the latest GemLib and a few other packages:
http://vincent.riviere.free.fr/soft/m68k-atari-mint/
http://vincent.riviere.free.fr/soft/m68k-atari-mint/ubuntu.php

However I completely agree with @th-otto:
> Is that really needed? It only removes some ifdefs, but even if vincent updates his toolchain, that does not mean that users immediately install his newer version.

Even though I've updated my packages, this doesn't imply that everyone has upgraded. Specially users of rarely updated hosts such as Mac, etc.

mikrosk:
There's no rush, we can keep this one open for some time. Maybe there will be other reason to update the toolchain in the meantime so the motivation to upgrade will be bigger.

Updating gemlib in your cross tools was long overdue anyway (there was already a build break in xaaes for similar reasons but we worked around it with a version check) so thank you for taking the time for doing it.

mikrosk:
Merged, year and a half seems like a long enough period for everyone to update.

PR #326, opened 2023-12-10
Contribute freemint Vampire V2 Expansion Port Ethernet Driver

infusixllc:


jvaltane:
There is a bug in v2expeth_recv(). Every time it receives any package MyAES hanged and sometimes machine rebooted.

got_size was introduced as pointer which was newer assigned anything else than 0. -> prpblems

Created pull request for infusixllc:s repository to fix the problem. https://github.com/infusixllc/freemint/pull/1

mikrosk:
This has been merged (incl. the linked PR) in https://github.com/freemint/freemint/commit/3725d69359c9125881764478d8fb6e46ca893dcb. I have also moved both `v2expeth` and `v4net` into separate subfolders so they don't increase the clutter in the `xif` folder.

@infusixllc @jvaltane which freemint kernel / package from https://freemint.github.io/#snapshots are you actually booting? I can create a separate folder (e.g. `vampire` ?) in `$SYSDIR` with those drivers pre-built for everyone (and even add Vampire's separate `MCHDIR` so the network driver would be loaded from that `vampire` subfolder automatically). Also, right now the drivers are compiled for `-m68000` so we can change that, too if desired.

PR #328, opened 2023-12-18
Fix regression introduced in #251

mikrosk:
It has been more than year when Lars [reported](https://sourceforge.net/p/freemint/mailman/freemint-discuss/thread/2120958839.99834.64094%40latzmaster.gmx.de/#msg37732330) sudden crashes on his CTPCI. In short, FreeMiNT + XaAES + other apps would work either with NVDI + MP off or without NVDI + MP on. What was quite perplexing as my fix of #251 clearly worked then (which was supposed to fix running FreeMiNT + XaAES without NVDI and MP on) and yet when reverting it, everything was fine again.

Over the year I exchanged many emails with Lars with test builds, debug outputs etc and I couldn't find it. On the top of that, Jo Even helped me to bisect the date when crashes on his Milan started to appear -- and surprise, surprise, it was the same commit as for Lars.

So I pulled my CTPCI out of the closet, set it all up and after a while and with some luck I was able to see what I hadn't see all that time -- that the original workaround with checking the PC against ROM space was still needed! So on one hand I did fix the problem with XaAES, TosWin2, Qed etc but on the other hand I removed the workaround which was essential for Milan and CTPCI VDI. During all those tests with Lars it never occured to me that this could be the culprit, I was seeing the strangest symptoms from the debug logs. Argh.

As a bonus, the first commit works around one annoying bug in CTPCI TOS: it doesn't allow specifying `device` = 1. So we have to ask for current resolution and pass that number to make it work.

Publishing for visibility, in case you see something dumb.

### DIFF
--- a/sys/dosmem.c
+++ b/sys/dosmem.c
@@ -178,19 +178,27 @@ sys_m_xalloc (long size, int mode)
 	TRACE(("Mxalloc(%ld,%x)",size,mode));
 
 #ifndef M68000
-	/* Hack here: if this is a Malloc call made by TOS 3/4 VDI's v_opnvwk routine,
-	 * we change mode to include "super accessible."
+	/* AKP: Hack here: if the calling process' PC is in ROM, then this
+	 * is a Malloc call made by VDI's v_opnvwk routine.  So we change
+	 * mode to include "super accessible."  This is temporary, until
+	 * VDI catches up with multitasking TOS.
+	 *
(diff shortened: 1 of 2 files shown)


th-otto:
Well i can't check anything CTPCI related, but i can't see anything obviously wrong there. So just go ahead, and congrats for finding such a nasty bug.

PR #335, opened 2024-04-17
mint060sp (motorola 060isp/fpsp)

agranlund:
The Motorola 060 support package emulates the missing Integer as well as Fpu instructions, thus:
- Removed complaints when using on LC/EC and have it install the integer part.
- Renamed the executable from fpu060sp -> 060sp for the same reason.

In addition:
- Fixed broken 68060 detection.
- Fixed a couple of callout functions.

Should this executable be built and distributed with the 060 Freemint builds?
CT60 TOS has the 060 isp/fpsp built in already but for other hardware running EmuTos this is handy.

Preferably, this should come with EmuTOS builds rather than Freemint since it's equally useful when running EmuTOS only.

### DIFF
--- a/sys/arch/060sp/Makefile
+++ b/sys/arch/060sp/Makefile
@@ -15,7 +15,7 @@ include $(top_srcdir)/CONFIGVARS
 include $(top_srcdir)/RULES
 include $(top_srcdir)/PHONY
 
-all-here: objs fpu060sp.prg
+all-here: objs 060sp.prg
 
 # default overwrites
 INCLUDES = -I$(top_srcdir)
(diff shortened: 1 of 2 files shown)


th-otto:
>     * Removed complaints when using on LC/EC and have it install the integer part.

I don't think this is the correct thing to do. The fpsp only emulates the instructions that were removed from 060 (like fsin etc), but not the ones that are still there. To support 060 without FPU, you would have to emulate much more.

mikrosk:
Are you sure @th-otto that you have interpreted this patch correctly? My understanding is that on LC/EC (no FPU) the FPSP wouldn't install at all, therefore omitting the important stuff as emulation of missing integer instruction which causes crash for some apps.

This patches enables emulation of integer instructions and keeps the rest as it was, i.e. it's clearly an improvement over the previous state (where *both* integer and FPU instructions were causing crashes).

> Preferably, this should come with EmuTOS builds rather than Freemint since it's equally useful when running EmuTOS only.

I tend to agree, it should not even "come with" but be "built in" EmuTOS, same as the PMMU initialisation for 040+.

mikrosk:
Merged, thanks!

Issue #343, opened 2024-07-19
Makefile problems


th-otto:
Something strange is going one when updating a sourcefile that is part of inet4.xdd. Some of the objects get recompiled, some not, even without running parallel makes:

```
-rwxr-xr-x 1 sebilla users  86418 19. Jul 16:42 ./sys/sockets/.compile_000/inet4.xdd
-rwxr-xr-x 1 sebilla users  83923 19. Jul 16:43 ./sys/sockets/.compile_02060/inet4.xdd
-rwxr-xr-x 1 sebilla users  84205 19. Jul 16:42 ./sys/sockets/.compile_030/inet4.xdd
-rwxr-xr-x 1 sebilla users  84249 19. Jul 16:43 ./sys/sockets/.compile_040/inet4.xdd
-rwxr-xr-x 1 sebilla users  83923 19. Jul 16:42 ./sys/sockets/.compile_060/inet4.xdd
-rwxr-xr-x 1 sebilla users  88263 19. Jul 16:42 ./sys/sockets/.compile_col/inet4.xdd
-rwxr-xr-x 1 sebilla users 103885 19. Jul 16:43 ./sys/sockets/.compile_deb/inet4.xdd
```

Any idea what might be causing this? It works when i delete the object files first, so maybe a VPATH issue?

Issue #32, opened 2017-05-07
Improvements in mint.cnf (xaaes.cnf) parsing


mikrosk:
I can think of two major improvements:

- allow usage of relative paths i.e. if mint.cnf is located in `c:\mint\mint-123\mint.cnf` it wouldn't complain about executing binaries relative to that (or $SYSDIR -- perhaps easier) path
- allow usage of either real env variables or some faked ones -- $SYSDIR and $MCHDIR are obvious candidates (this reminds me introduction of $MCHDIR as an env variable...)

With at least one of the changes one could use universal configuration files across different builds, esp. true for `INIT`/`GEM` and `xaloader.prg` path.

paulwratt:
this is definitely a candidate for a v1.20 release.

note that ENV parsing should be on per line basis, not per file, as both MiNT and XaAES define ENV variables through out there configs.

th-otto:
Environment variable parsing has now been implemented in https://github.com/freemint/freemint/commit/8f23fa118809e9f5dfec0d572a8250e9d0e7e3d4. Only real environment variables are supported (ones that are set using setenv, not any local ones that are just assigned as in `KERN_DEBUG_LEVEL=1`). 

`$SYSDIR` was already set in https://github.com/freemint/freemint/blob/8f23fa118809e9f5dfec0d572a8250e9d0e7e3d4/sys/init.c#L1080, i've also added a few others like `$MCHDIR`, `$MINT_VERSION`, and `$BOOTDRIVE`

There is one thing that may behave differently than before: in the previous version, all need conversions (like stripping quotes, convert escape sequences) was done to the same buffer the one the file contents were read to. That worked because it would never write more bytes, only fewer. When implementing this, i had to use fixed line buffers instead. That means, input lines are now silently truncated if they are too long. The size is currently set to 512, so i think this is not a real problem in practise.

The doc/example/mint.cnf has already been adjusted accordingly, and the sysdir tool should no longer be needed (it is still build and installed though, for the the per-cpu builds, where mint.cnf is not updated).

mikrosk:
Nice! 

I just wish you separated whitespace changes from the actual implementation, now the diffs are huge and very hard to review. :-/

>  (it is still build and installed though, for the the per-cpu builds, where mint.cnf is not updated)

[...]

PR #348, opened 2024-07-28
Rework makefiles

th-otto:
I'm currently working of reworking the Makefiles for freemint, hopefully making them simpler and more stable.

Main reason to start this was https://github.com/freemint/freemint/issues/343, but i occasionally also have problem when running parallel makes.

This is still work in progress, more commits will be pushed to this branch, but maybe you can already have a look whether there are any objections.

Things that i noticed in the meantime:

- isofs.xfs was only compiled for m68000 until now, and not distributed by util.sh
- minix.xfs attempts to keep some specific symbols (from syms.mfs) when stripping. I missed this in the ELF adjustments (they need the undercores removed for elf), but does anybody know for what they are needed?


mikrosk:
So far it looks good to me but frankly, it is equally cryptic as it was before. ;)

So what are you high-level goals with this, i.e. what do you want to reduce/remove?

As for  isofs.xfs my understanding was that it is just a skeleton / early WIP by Frank.

th-otto:
I admit, sometimes the templates are a bit cryptic. Especially since you have to quote Makefile special variable like `$<` most the time (i guess this is because of the `$(eval)` that has to be used). 
But it makes it much easier to extend/reduce the targets that drivers are compiled for (audiodev.xdd for example was only compiled for 000 until now). It also makes them more consistent, eg. that the `deb` debug target is compiled for 020-60 is now only specified once, in the toplevel rules, not in a lot of *DEFS files.
The pattern rules should also be obsolete once everything is finished, these sometimes work different depending on what make version you use.
All in all it saves also a lot of recursions, since there is now only one Makefile in the subdirectories, not duplicate Makefile/Makefile.objs with repeated but different settings. And that VPATH thing and/or the pattern rules do not always work.

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
 	mkdir -p "$MCHDIR"
 	# TODO: perhaps these four are compatible also with other machines
 	#       but they are awfully old and obsolete so let's keep them here
-	cp "$SRC/sys/sockets/xif/biodma.xif" "$MCHDIR/biodma.xix"
-	cp "$SRC/sys/sockets/xif/de600.xif" "$MCHDIR/de600.xix"
-	cp "$SRC/sys/sockets/xif/dial.xif" "$MCHDIR/dial.xix"
-	cp "$SRC/sys/sockets/xif/pamsdma.xif" "$MCHDIR/pamsdma.xix"
-	cp "$SRC/sys/sockets/xif/rtl8012st.xif" "$MCHDIR/rtl8012st.xix"
-	cp "$SRC/sys/xdd/mfp/mfp.xdd" "$MCHDIR"
+	cp "$SRC/sys/sockets/xif/biodma_000.xif" "$MCHDIR/biodma_000.xix"
```
mikrosk, line 61: Is there any benefit of keeping the _000 suffix there? (it also breaks 8+3 filename limit)

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
 	local MCHDIR="$1/tt"
+	local TARGET="$2"
 	mkdir -p "$MCHDIR"
-	cp "$SRC/sys/sockets/xif/daynaport/scsilink.xif" "$MCHDIR/scsilink.xix"
-	cp "$SRC/sys/sockets/xif/lance.xif" "$MCHDIR/lance.xix"
-	cp "$SRC/sys/sockets/xif/riebltt.xif" "$MCHDIR/riebltt.xix"
-	cp "$SRC/sys/sockets/xif/rtl8012.xif" "$MCHDIR/rtl8012.xix"
-	cp "$SRC/sys/xdd/mfp/mfp.xdd" "$MCHDIR"
-	cp "$SRC/sys/xdd/scc/scc.xdd" "$MCHDIR"
+	cp "$SRC/sys/sockets/xif/daynaport/.compile_${TARGET}/scsilink.xif" "$MCHDIR/scsilink.xix"
```
mikrosk, line 97: Why is now ${TARGET} used and not hardcoded _030 (similar to _000 above) ? TT can't be run on other than 030 CPU.

.scripts/util.sh
```diff
@@ -126,25 +129,26 @@ copy_ct60_modules() {
...
-	cp "$SRC/sys/xdd/nfstderr/nfstderr.xdd" "$MCHDIR"
-	cp "$SRC/sys/xdd/nfexec/nfexec.xdd" "$MCHDIR"
+	cp "$SRC/sys/sockets/xif/nfeth/.compile_040/nfeth.xif" "$MCHDIR"
+	cp "$SRC/sys/xdd/nfstderr/.compile_040/nfstderr.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/nfexec/.compile_040/nfexec.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/nfcdrom/.compile_040/nfcdrom.xdd" "$MCHDIR/nfcdrom.xdx"
 	cp "$SRC/sys/xfs/aranym/aranym.xfs" "$MCHDIR"
 	# not really needed
-	cp "$SRC/sys/xfs/hostfs/hostfs.xfs" "$MCHDIR/hostfs.xfx"
+	cp "$SRC/sys/xfs/hostfs/.compile_000/hostfs.xfs" "$MCHDIR/hostfs.xfx"
```
mikrosk, line 151: Why _000 and not _040?

RULES
```diff
@@ -7,23 +7,57 @@ include $(srcdir)/SRCFILES
...
+MODEL_02060 = -m68020-60
+MODEL_030 = -m68030
+MODEL_040 = -m68040
+MODEL_060 = -m68060
+MODEL_v4e = -mcpu=5475
+
+#
+# mapping of TARGET names to CPU
+#
+CPU_deb = 02060
```
mikrosk, line 24: Wouldn't make sense to make this more explicit and map them to the MODEL names?

sys/KERNELDEFS
```diff
@@ -46,109 +45,69 @@
...
+#KERNELDEFS_my1 += -DPROFILING
+# add this to ../RULES
+CPU_my1  = 02060
 
 #
 # all default targets
 #
 kerneltargets = 000 020 030 040 060 080 deb mil ara hat hat030 col
-cputargets = 000 020 030 040 060 020-60 v4e
+cputargets = 000 020 030 040 060 02060 v4e
```
mikrosk, line 113: Since you have been working on this... isn't there a way to get rid of this duplicity? To this day I'm always confused when I should enable/disable entries in `kerneltargets` and `cputargets`. Having just `kerneltargets` would make so much more sense.

sys/sockets/xif/nfeth/Makefile
```diff
@@ -1,47 +1,74 @@
...
+all-here: $(DRIVERS)
+
+strip:
+	$(STRIP) $(DRIVERS)
+
+#
+# multi target stuff
+#
+
+LD = $(CC) -s -nostdlib -Wl,--entry,init -Wl,--fatal-warnings
```
mikrosk, line 43: Again the forced `-s` ...

sys/xdd/nfexec/Makefile
```diff
@@ -12,29 +10,84 @@ subdir = nfexec
...
+strip:
+	$(STRIP) $(ALL_TARGETS)
+
+all-targets: $(ALL_TARGETS)
+
+#
+# multi target stuff
+#
+
+LD = $(CC) -s -nostdlib -Wl,--entry,init_xdd -Wl,--fatal-warnings
```
mikrosk, line 62: Again the forced `-s` ...

sys/xfs/aranym/Makefile
```diff
@@ -12,29 +10,67 @@ subdir = aranym
...
+all_compile_dirs = .compile_*
+GENFILES = $(all_compile_dirs) $(TARGET)
+
+all-here: $(TARGET)
 
 strip:
 	$(STRIP) $(TARGET)
 
-ifneq ($(kernel),)
+LD = $(CC) -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings -s
```
mikrosk, line 43: Again the forced `-s` ...

sys/xfs/ext2fs/Makefile
```diff
@@ -41,46 +48,47 @@ help:
...
+	$(STRIP) $(ALL_TARGETS)
+
+all-targets: $(ALL_TARGETS)
 
 #
 # multi target stuff
 #
 
-ifneq ($(ext2),)
+LD = -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings -s
```
mikrosk, line 63: Again the forced `-s` ...

sys/xfs/hostfs/Makefile
```diff
@@ -12,29 +25,65 @@ subdir = hostfs
...
+targets = 000 ara
+
+all-here: $(targets)
 
 strip:
-	$(STRIP) $(TARGET)
+	$(STRIP) .compile_*/*.xfs
 
-ifneq ($(kernel),)
+LD = $(CC) -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings -s
```
mikrosk, line 55: Again the forced `-s` ...

sys/xfs/isofs/Makefile
```diff
@@ -12,27 +11,67 @@ subdir =
...
+strip:
+	$(STRIP) $(ALL_TARGETS)
+
+all-targets: $(ALL_TARGETS)
+
+#
+# multi target stuff
+#
+
+LD = $(CC) -s -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings
```
mikrosk, line 46: Again the forced `-s` ...

sys/xfs/minixfs/Makefile
```diff
@@ -39,46 +45,47 @@ help:
...
+ALL_TARGETS = $(addsuffix /minix.xfs, $(addprefix .compile_,$(minixtargets)))
+
+all-targets: $(ALL_TARGETS)
 
 #
 # multi target stuff
 #
 
-ifneq ($(minix),)
+LD = $(CC) -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings -s
```
mikrosk, line 60: Again the forced `-s` ...

sys/xfs/nfs/Makefile
```diff
@@ -39,46 +41,47 @@ help:
...
+	$(STRIP) $(ALL_TARGETS)
+
+all-targets: $(ALL_TARGETS)
 
 #
 # multi target stuff
 #
 
-ifneq ($(nfs),)
+LD = -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings -s
```
mikrosk, line 56: Again the forced `-s` ...

sys/xfs/skeleton/Makefile
```diff
@@ -12,27 +10,67 @@ subdir =
...
+strip:
+	$(STRIP) $(ALL_TARGETS)
+
+all-targets: $(ALL_TARGETS)
+
+#
+# multi target stuff
+#
+
+LD = $(CC) -s -nostdlib -Wl,--entry,init_xfs -Wl,--fatal-warnings
```
mikrosk, line 45: Again the forced `-s` ...

mikrosk (commented):
Overall it's definitely an improvement. I'm a bit scared of all those $(1) stuff but compared to the old version it's still much better readable.

Just please don't enforce those `-s` flags, if `make strip` doesn't work, that has to be fixed. Debug symbols are useful out of the box.

th-otto:
> Overall it's definitely an improvement. I'm a bit scared of all those $(1) stuff but compared to the old version it's still much better readable.

Yes, i think too. Maybe some more stuff can later be moved to included files (and then better documented), but for now it should do.

> Debug symbols are useful out of the box.

I still don't see this, especially not for modules that are loaded by the kernel or the usb loader. But well.

.scripts/util.sh
th-otto: No, just a copy/paste error.  Will fix it.

.scripts/util.sh
th-otto: Because it is called with different CPU targets (030 for tt, 02060 for falcon)

.scripts/util.sh
th-otto: Because the standalone module is compiled for 000. For the aranym kernel, that filesystem is compiled into the kernel anyway. The standalone module would only be useful for other CPU targets, but then you can of course not boot from hostfs (but theoretically it could still be used to access other drives)

RULES
th-otto: Currently that does not work. CPU names are used in the libkern subdirectory (and a few other places), whereas most of the other Makefiles use kernel target names like ara, hat etc. So unfortunately that has to be mapped twice.

sys/KERNELDEFS
th-otto: See above, we need both, because there is no libkernhat.a. Some subdirectories (like usb) also need extra target names like prg, that are not in the toplevel `RULES`

.scripts/util.sh
mikrosk: True but those are copied into the MCHDIR, i.e. being used just on that machine. In this case the TT.

When I introduced the 02060 target, the idea was to offer a package which includes these CPU families (i.e. one archive to run on 020, 030, 040, 060), not to explicitly compile everything for that CPU (of course, it makes sense for the tools and common code because you don't know on which machine it will end up running).

So it's totally fine to keep `copy_tt_modules` as copying, well, modules aimed for the TT, i.e. 030 CPU.

.scripts/util.sh
mikrosk: OK, fair point. However then it makes no sense to put it into Aranym's MCHDIR because that implies it is usable only for the aranym kernel.

RULES
mikrosk: Maybe it's worth mentioning this directly in the Makefile then.

mikrosk:
> I still don't see this, especially not for modules that are loaded by the kernel or the usb loader. But well.

If for nothing else: at least it would be consistent. No "make strip" => everything has symbols. Currently some of them have symbols, some don't. IIRC it used to be like this in the past, too but perhaps the broken `make strip` targets (which were missing many of the built binaries) forced people to add those `-s` switches.

.scripts/util.sh
th-otto: Yes, but it currently does not make much difference anyway. There is only one archive for tt-falcon, not separate ones, so these modules will be overwritten later by copy_falcon_modules.

th-otto:
One problem of the `make strip` target is, that you have to compile *all* kernels first, otherwise it will fail because of missing executables. Often if do just `make clean; make ara`, and that will leave then unstripped binaries around.

.scripts/util.sh
mikrosk: That's my point, they won't -- cooy_falcon_modules will copy stuff into Falcon MCHDIR so there's really no point in compiling TT modules for other than 030.

th-otto: But this already happens in other places too: https://github.com/freemint/freemint/actions/runs/10158122454/job/28089528657?pr=348#step:8:5308

And IIRC, i got some errors from cp command when i omit 02060

th-otto:
We could also use something like https://github.com/freemint/freemint/blob/b227ddf9df1513b6fbf9b3b85779e5e917dbdb94/sys/usb/src.km/ucd/vttusb/Makefile.objs#L68-L70 consistently in all Makefiles (by adding `-s` only if `NOSTRIP` is not set)

.scripts/util.sh
mikrosk: usb modules are OK to compile for 02060: that's the common code I was referring to. With those, you can't tell whether the zip will be unpacked/booted on the TT or Falcon, so they are 020-60. However TT, Falcon, Aranym MCHDIRs are solely for given machine, there's no benefit in compiling them for something else.

Actually your work here is very beneficial: until now we didn't have an option to compile, say, mfp.xdd for 030 and for 040 and now we can do it per-MCHDIR: Falcon's and TT's MCHDIR would contain 030 mfp.xdd, Aranym 040 mfp.xdd, Hades 020-60 mfp.xdd ... very cool.

So I think this is worth looking into, TT should be definitely 030 only.

th-otto: Yes, i've changed that copy_tt_modules already to use 030, not `$TARGET` (not pushed yet). I first thought those modules would be overwritten by copy_falcon_modules, until i realized that they use different directories.

What should be checked also is copy_m68k_modules where i did a similar change. Until now some of these modules were only compiled for `000`.

th-otto:
Just noticed: https://github.com/freemint/freemint/blob/b227ddf9df1513b6fbf9b3b85779e5e917dbdb94/tools/fsetter/Makefile.objs#L29-L32

Are any libraries installed to /usr/GEM? That won't work because -lcflib will not lookup architecture dependent libraries there.

mikrosk:
> Are any libraries installed to /usr/GEM? That won't work because -lcflib will not lookup architecture dependent libraries there.

I'm not aware of any, it's an ancient Makefile. Perhaps back then nobody cared about 020+ GEM applications.

th-otto:
Not a big issue for 020+, but fatal for coldfire...

.scripts/util.sh
mikrosk: On the contrary, the `copy_m68k_modules` change is exactly what we want: each target will get optimised modules. In case of the bootable TT/Falcon archive, we get 020-60 modules instead of 000 modules => hopefully better performance.

There used to be an issue when actually USB modules were compiled for 000 only and tested on real (030/060) hardware and after changing the CPU flags timing didn't match anymore. However that was fixed by @DavidGZ a long time ago. Hopefully other modules (especially the new ones: asix, picowifi) don't use constructs like "move #random_constant, d0; dbra d0, loop" for delays.

mikrosk:
I think it's safe to delete, it has no real value these days. Maybe some ancient CFLib RPM installed stuff there.

th-otto:
> I think it's safe to delete, it has no real value these days. Maybe some ancient CFLib RPM installed stuff there.

Yes, i think too. Have removed the -I and -L directives, lets see how it goes when running on github.

mikrosk:
Btw Thorsten, any chance you could include as part of your rework also building of https://github.com/freemint/freemint/tree/master/sys/arch/060sp ? We could supply it with the 060 kernels.

th-otto:
Maybe as separate project later, but it is already available at https://github.com/th-otto/fpsp.

Main issue was that the original sources are not compilable by gas (they used sun as, which has some pitfalls like reversed operands in the cmp instruction)

th-otto:
Just noticed another strangeness:
https://github.com/freemint/freemint/actions/runs/10120442540/job/27990204801#step:8:5166

LIBCMINI_INCPATH is passed when linking, but not when compiling. That's a bit dangerous, since especially the `<stdio.h>` is completely different.

That effects nohog2, the new sysdir utility and usbtool.acc.

mikrosk:
> Main issue was that the original sources are not compilable by gas (they used sun as, which has some pitfalls like reversed operands in the cmp instruction)

Is this is the case also with FreeMiNT sources? @agranlund compiled them just fine here: https://github.com/agranlund/raven/tree/main/sw/support. But you are right, your version is much more polished so we should deliver that one (same as with QED etc) but it's missing the EC/LC support Anders made a few months ago: https://github.com/freemint/freemint/commit/3a4363c4242a66f6d20d3f4eba3a757be527028c.

> LIBCMINI_INCPATH is passed when linking, but not when compiling

Oh dear. That certainly wasn't intentional.

> it took 15min20 from the top-level, with the new version, it took 11min28

Imagine how much it would have saved on plain Falcon030 :-)

th-otto:
On aranym, `make 000` takes about 24min, but that only includes the kernel and xaaes, not the drivers, tools, etc.

th-otto:
Phew. I think that should be it for the moment.

Some things could maybe be moved to separate include files, but i'm not sure whether that improves anything, there are a lots of exceptions everywhere (ie. in sys/xdd/audio also an executable is compiled, which must not use mshort)

Let me know if everybody thinks this should be merged.

PS.: the last commit should finally solve your strip problem. NOSTRIP can now be used for that purpose.

th-otto:
> @agranlund compiled them just fine here: https://github.com/agranlund/raven/tree/main/sw/support. 

Yes, but that uses only the already compiled *.sa files, which look like
```
	dc.l	$60ff0000,$238e0000,$60ff0000,$24200000
	dc.l	$60ff0000,$24b60000,$60ff0000,$11060000
	dc.l	$60ff0000,$11980000,$60ff0000,$122e0000
```

And even that is fed through the funny asm2gas sed script.

And as far as i can see, he only changed the _CPU detection using the cookie jar. The actual EC/LC detection is the same, only the error messages are removed.

Apart from that, if we want to support the binaries, we would also need the fpsp for 040.

agranlund:
I can’t recall exactly what I changed but it wasn’t much.
It used to bail out with an error immediatly on LC’s and the change was to let it install the Integer emulation package and then exit (without installing the floating point package, and without displaying the error message in this case since ignoring the fpu part is intended behavior on LCs)

th-otto:
If i look at the original code (https://github.com/freemint/freemint/blame/d0d65b21285d788c1769396454cb6709f2fc4c2d/sys/arch/060sp/mint060sp.S#L82-L107), it prints an error message, but then continues to install the integer emulation package. So the only real difference is the printing of the message.

But i wonder whether it makes sense to integrate that into the kernel somehow. You may as well just execute the standalone version from mint.cnf. So all that needs to be changed is to build it, and make it available in the archives.

th-otto:
Any objections to merge this? I would like to have the snapshots been generated by the new system, so that we can see whether there are any issues.

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
-	cp "$SRC/sys/xdd/mfp/mfp.xdd" "$MCHDIR"
-	cp "$SRC/sys/xdd/scc/scc.xdd" "$MCHDIR"
+	cp "$SRC/sys/sockets/xif/rieblste_000.xif" "$MCHDIR/rieblste.xix"
+	cp "$SRC/sys/sockets/xif/rtl8012st_000.xif" "$MCHDIR/rtl8012st.xix"
+	cp "$SRC/sys/xdd/mfp/.compile_000/mfp.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/scc/.compile_000/scc.xdd" "$MCHDIR"
 }
 copy_tt_modules() {
 	local MCHDIR="$1/tt"
+	local TARGET="$2"
```
mikrosk, line 95: This is now obsolete (also this function is still called with the CPU argument)

.scripts/util.sh
```diff
@@ -252,8 +265,9 @@ copy_ehci_usb_modules() {
 
 copy_ct60_usb_modules() {
 	local USBDIR="$1"
+	local TARGET="$2"
```
mikrosk, line 268: This is not needed. CT60 (and therefore Ethernat) always implies 68060. It just adds confusion why *CT60* function has to be called with "060".

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
+	cp "$SRC/sys/xdd/scc/.compile_030/scc.xdd" "$MCHDIR"
 }
 copy_falcon_modules() {
 	local MCHDIR="$1/falcon"
 	mkdir -p "$MCHDIR"
-	cp "$SRC/sys/sockets/xif/rtl8012.xif" "$MCHDIR/rtl8012.xix"
-	cp "$SRC/sys/sockets/xif/daynaport/scsilink.xif" "$MCHDIR/scsilink.xix"
-	cp "$SRC/sys/xdd/dsp56k/dsp56k.xdd" "$MCHDIR"
-	cp "$SRC/sys/xdd/scc/scc.xdd" "$MCHDIR"
+	cp "$SRC/sys/sockets/xif/rtl8012_02060.xif" "$MCHDIR/rtl8012.xix"
```
mikrosk, line 107: While this is not wrong, it could be done in a better way, as discussed in the other thread:

- for `freemint-latest-02060-tt_falcon_clones.zip` 02060 is great (as it can be booted on both 030 and 040 Falcons; the AB040 has no separate `MCHDIR` ... which maybe should)
- for `freemint-latest-030.zip` and `freemint-latest-040.zip` there's no reason why the drivers shouldn't be compiled directly for given CPU

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
+	cp "$SRC/sys/sockets/xif/daynaport/.compile_02060/scsilink.xif" "$MCHDIR/scsilink.xix"
+	cp "$SRC/sys/xdd/dsp56k/.compile_02060/dsp56k.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/scc/.compile_02060/scc.xdd" "$MCHDIR"
 }
 copy_milan_modules() {
 	local MCHDIR="$1/milan"
 	mkdir -p "$MCHDIR"
-	cp "$SRC/sys/xdd/uart/uart.xdd" "$MCHDIR"
-	cp "$SRC/sys/xdd/mfp/mfp_mil.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/uart/.compile_mil/uart.xdd" "$MCHDIR"
```
mikrosk, line 115: See above, while mil = 02060 is good for the generic bootable archive, it could use 040 and 060 for the CPU builds.

.scripts/util.sh
```diff
@@ -38,85 +38,88 @@ copy_modules() {
...
+	cp "$SRC/sys/xdd/mfp/.compile_mil/mfp.xdd" "$MCHDIR"
 }
 copy_hades_modules() {
 	local MCHDIR="$1/hades"
 	mkdir -p "$MCHDIR"
 	# Hades ESCC is compatible with TT SCC
-	cp "$SRC/sys/xdd/scc/scc.xdd" "$MCHDIR"
+	cp "$SRC/sys/xdd/scc/.compile_02060/scc.xdd" "$MCHDIR"
```
mikrosk, line 122: See above, while 02060 is good for the generic bootable archive, it could use 040 and 060 for the CPU builds.

mikrosk:
> You may as well just execute the standalone version from mint.cnf. So all that needs to be changed is to build it, and make it available in the archives.

I think that's more than enough to make users (and @agranlund ;)) happy.

> Any objections to merge this?

I have posted a few more comments in the review. After resolving them (easy stuff) I don't have any objections.

I'm thinking that maybe you should keep your commits as they are, i.e. not squash them. Each commit is not buildable (I presume) but it offers a non-overwhelming way to review your changes step by step.

.scripts/util.sh
th-otto: Fixed.

.scripts/util.sh
mikrosk: Well, the reason for existence of the CPU builds is itself very thin with the `sysdir.tos` and #32 being implemented (my initial idea was to provide them for a) stubborn users requiring their CPU be utilised b) users which just wish to extract&overwrite their existing installation).

So making every single module optimised for the given CPU seems like a nice feature. Otherwise we may really consider killing the whole "CPU builds" concept.

.scripts/util.sh
mikrosk: Huh. I haven't realised the effect of `deb) TEMP_CPU_TARGET=02060` until now.

In that case this is still inconsistent with other machine-specific modules, isn't it? E.g. for the TT we always use `030`, so why suddenly `${TARGET}` here?

I think `deb` target shouldn't change / override CPU definitions, it's quite confusing. Maybe it could be transformed into a flag, i.e. if building with `DEBUG=1`, add proper `CFLAGS` etc.

th-otto:
> Each commit is not buildable (I presume) 

Actually they should be. Maybe there are some issues in the github scripts that are fixed in later commits, but they should compile atleast.

.scripts/util.sh
th-otto: That won't solve the problem. You need to use a different .compile_dir when compiling the same source with different settings (.compile_deb in this case).

Using the target to override different settings is also used in other places, mainly the sockets/xif directory where the lance driver is compiled for st/mste etc.

mikrosk: Ok but that still doesn't answer my question why did you choose this specific function to and not others, where the CPU is hardcoded as well?

mikrosk:
> I have posted a few more comments in the review (plus the couple of old/unresolved yet ones). After resolving them (easy stuff) I don't have any objections.

I would prefer if you merged this _after_ we resolve all the threads in 
image

.scripts/util.sh
th-otto: Because the deb target is not used in all subdirectories so far.

The deb target is a bit of a hack anyway. Until now it was not even build by the scripts (see also the commit for the scc driver where i had to fix some printfs). It is also not deployed. And in some directories, the CPU type for it is changed to 030. I kept it this way because i wasn't sure about the reason for it.

th-otto:
> I would prefer if you merged this _after_ we resolve all the threads

Yes, maybe. But we can also resolve them in the master branch, should there be any issues that needs fixing. It doesn't make much difference. I wanted to merge them to avoid conflicts with other changes like the ones for the fpsp that are not directly related to the rework, but touch the same files.

.scripts/util.sh
mikrosk: OK, I create a new issue where this can be discussed.

PR #349, opened 2024-07-29
Fix a null pointer crash

### LINKED ISSUE #346: wind_update: single key press leads to FATAL ERROR
When playing with `menu_bar` I have discovered an interesting bug in XaAES. Since it was very tricky to reproduce in Hatari, I'm attaching the complete archive where I'm able to see the issue. It's the (currently) latest bootable snapshot (1-19-c4e) with `mint030h.prg` renamed to `mint030.prg` and removed useless folders/programs. Everything else is default. The issue was discovered on my TT030 hence I'm using TOS 3.06 but I have reproduced the issue also with EmuTOS 1.3 512K, so it's thankfully TOS-agnostic.

1. Unzip [wind_update.zip](https://github.com/user-attachments/files/16396639/wind_update.zip), go to `wind_update` folder.
2. Run `hatari -c ../hatari-tt.cfg --fast-forward 1 --conout 2 .` (most likely you don't need my Hatari config, attaching just to be on the safe side)
3. Open C:, disable fast forward (AltGr +X)
4. Right click on test.app
5. Wait until the screen starts flashing
6. Press any key, e.g. space, once is enough
7. Observe a total lockup.

This lockup can be avoided in at least two ways:
- disable the `menu_bar` calls
- disable the `wind_update` calls

Hard to say what is the nature of the problem here. I can't reproduce this on Aranym+FVDI, Falcon060+NVDI **but** I _can_ reproduce it on my Falcon030+NVDI (basically the same setup as on 060). So most likely this is some kind of race condition where on faster machines something (the menu redraw?) is executed fast enough but on my TT/Falcon030 isn't.

When playing in Hatari, I can observe the same kind of random behaviour: for instance if I remove all kernel modules _and_ move `mint030h.prg` into `../../auto/mint-c4e.prg`), I couldn't get it to crash so easily anymore.

XA_BOOT.LOG sometimes gives interesting hints, e.g. `exception 4 for AESSYS pc:1002238 addr:0` or `XaAES:../rectlist.c,402:freeing invalid memory: 107C308(-16777200/200641)` so it's clear the memory must be in a pretty bad shape.

Source code of `test.app` is present in the archive as well, I'm not aware of any wrong-doing there but feel free to correct me. The menu structure is copy&pasted from Resource Master 3.65 'C source' output, it's the default menu it offers.

mikrosk:
This fixes a surprisingly rare crash when S.focus == NULL. It's possible that it breaks some scenario:

- commit 0978888e explicitly changed `top->owner` to `S.focus->owner` which is really suspicious because not only `top->keypress` check was kept but also the body still had the original condition (`top->owner == client && top->keypress`)

- commit 62947f1e extended the body with `S.focus && S.focus->owner == client`, perhaps to address the inconsistency but didn't add a NULL check and also kept `top->owner == client && top->keypress` intact.

This leads me to believe that actually both conditions are valid and the change from `top->owner` to `S.focus->owner` was done in a hurry without thinking it through.

Fixes #346.

===

As you can see, in the end it was very obvious but I was totally mislead by the debug outputs. As mentioned in the commit message, I'm unsure what is the intended behaviour in the `if`, I went with most logical assumption but I can be wrong.

@th-otto @Landemarre @mfro0 @lpgb any input appreciated, this GEM stuff is just giving me headache. At least the crash is fixed.

### DIFF
--- a/xaaes/src.km/app_man.c
+++ b/xaaes/src.km/app_man.c
@@ -366,9 +366,10 @@ find_focus(bool withlocks, bool *waiting, struct xa_client **locked_client, stru
 			 * should not be taken into account when client wants TOP_WINDOW via wind_get(WF_TOP)!
 			 * But the AES may perhaps need to direct some keypresses to those later on..
 			 */
-			    (nlwind && nlwind->owner == client && !(nlwind->dial & created_for_POPUP)) ||
-			    client->waiting_for & (MU_KEYBD | MU_NORM_KEYBD) ||
-			    (S.focus->owner == client /*top->owner == client*/ && top->keypress))		/* Windowed form_do() */
+				(nlwind && nlwind->owner == client && !(nlwind->dial & created_for_POPUP)) ||
+				(client->waiting_for & (MU_KEYBD | MU_NORM_KEYBD)) ||
+				(top->owner == client && top->keypress) ||		/* Windowed form_do() */
+				(S.focus && S.focus->owner == client))
 			{
 				if (waiting)
 					*waiting = true;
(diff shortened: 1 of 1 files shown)


th-otto:
Hm, have to check, but according to other comments the problem is that AES sometimes has to send key input to popups, but these are not the "top window" (as reported by wind_get(WF_TOP)), even if they are on top of all other windows.

mikrosk:
Maybe that's the key difference between `S.focus` and `top`. In any case, my change shouldn't break this, just handle both (`top->owner` and `S.focus->owner`) in the same manner.

mikrosk:
I have been using this for the whole weekend, didn't notice anything weird so merging.

Issue #363, opened 2024-08-14
Does iflink ever worked?


MedourMehdi:
Hi,

Obviously the _[iflink](https://github.com/freemint/freemint/blob/master/tools/net-tools/iflink.c#L86)_ binary provided can't work.
It calls _[SIOCGIFNAME](https://github.com/freemint/freemint/blob/master/tools/net-tools/iflink.c#L86)_ who do nothing except check for permissions and then lead to an error.
It does bad usage of _[SIOCGIFNAME](https://github.com/freemint/freemint/blob/master/tools/net-tools/iflink.c#L61)_ with a wrong iflink structure and then lead to an error.

Does anyone know it ever worked? May be I misunderstand what is it for.

th-otto:
What make you think so? The call should end up in https://github.com/freemint/freemint/blob/0c8bc27e427d5b2097b5db07d185ca5cdffa44e2/sys/sockets/inet4/if.c#L590 , which then calls the ioctl of the driver to get the devicename. Only problem seems to be that `SIOCGIFNAME` does not seem to be implemented in most drivers, except for ppp (https://github.com/freemint/freemint/blob/0c8bc27e427d5b2097b5db07d185ca5cdffa44e2/sys/sockets/xif/ppp.c#L324) and slip (https://github.com/freemint/freemint/blob/0c8bc27e427d5b2097b5db07d185ca5cdffa44e2/sys/sockets/xif/slip.c#L217)

MedourMehdi:
Agreed but this isn't the role of SIOCGIFNAME as It should return an index.
Do we have to deal with sizeof(arg) to know if we should call the ioctl of the driver?

MedourMehdi:
What do you think about filtering with

if(( nif->hwtype & HWTYPE_PPP ) || ( nif->hwtype & HWTYPE_SLIP )){

}

I don't know how to deal with that.

th-otto:
Most simply, use a new define to map an index to interface name. That means also changes to your PR in mintlib.

MedourMehdi:
I've made some changes using another define (SIOCGIFNAME_ETH), I can change the name if you want.

MedourMehdi:
> Most simply, use a new define to map an index to interface name. That means also changes to your PR in mintlib.
Hi,

What do you think about using this in mintlib sockios.h ?
`#ifdef __USE_GNU

#ifdef SIOCGIFNAME
#undef SIOCGIFNAME
[...]

Issue #359, opened 2024-08-13
SIOCGIFNAME seems wrong / SIOCGIFINDEX implementation


MedourMehdi:
In order to implement if_indextoname & if_nametoindex in mintlib I've added the following code in if_ioctl function

```
switch (cmd)
	{
		case SIOCGIFINDEX:
		{
			ifr->ifru.ifindex = if_name2index(ifr->ifr_name);
			if(ifr->ifru.ifindex != -1){
				return 0;
			}
			DEBUG (("if_ioctl: %s: no such if", ifr->ifr_name));
			return ENOENT;
		}
	}
```

with

```
short 
if_name2index (char	*ifr_name)
{
	struct netif *ifp;
	
	for (ifp = allinterfaces; ifp; ifp = ifp->next)
	{
		if (!strncmp (ifp->name, ifr_name, IF_NAMSIZ)){
			return ifp->unit;
		}
	}
	return -1;
}
```

Then I figured out that _short   ifindex;_ was missing from _struct ifreq_

So I wondered how _SIOCGIFNAME_ could work If there isn't any ifindex?

From https://man7.org/linux/man-pages/man7/netdevice.7.html 
```
       SIOCGIFNAME
              Given the ifr_ifindex, return the name of the interface in
              ifr_name.  This is the only ioctl which returns its result
              in ifr_name.

```
Fyi

```
char *if_indextoname(unsigned index, char *name)
{
	struct ifreq ifr;
	int fd, r;

	if ((fd = socket(AF_INET, SOCK_DGRAM, 0)) < 0) return 0;
	ifr.ifr_ifindex = index;
	r = ioctl(fd, SIOCGIFNAME, &ifr);
	close(fd);
	return r < 0 ? 0 : strncpy(name, ifr.ifr_name, IF_NAMSIZ);
}

unsigned int if_nametoindex(const char *name)
{
	struct ifreq ifr;
	int fd, r;

	if ((fd = socket(AF_INET, SOCK_DGRAM, 0)) < 0) return -1;
	strlcpy(ifr.ifr_name, name, sizeof ifr.ifr_name);
	r = ioctl(fd, SIOCGIFINDEX, &ifr);
	close(fd);
	return r < 0 ? r : ifr.ifr_ifindex;
}
```

th-otto:
> So I wondered how _SIOCGIFNAME_ could work If there isn't any ifindex?

[...]

PR #365, opened 2024-08-14
siocgifname_eth & siocgifindex calls implementation

MedourMehdi:
Linux netdevice compatible calls for siocgifname & siocgifindex.

siocgifname_eth is used indtead of siocgifname (in order to keep compatibility with code using siocgifname call with ppp & slip interface).

cf. #359 & #363


MedourMehdi:
Check [comment here](https://github.com/freemint/freemint/pull/412#issuecomment-2993042854)

PR #379, opened 2024-09-14
Partially revert commit 9668d67

### LINKED ISSUE #143: 1.19.x kernels require gemcon=false in N.AES config file
N.AES do not longer work together with the MiNT Kernel 1-19. Keyboard inputs (press a key) not longer work. Mouse works fine. With the 1-18-0 Kernel all works ok.

### LINKED ISSUE #173: Compatibility with AES4.1?
Hi guys,

Sorry if this is an old topic, but couldn't see anything about it in the issues list.

I like XaAES and it seems very stable, but I also like the AES3/4 desktop rather than any third party offering. I spend half my time in MiNT and half in vanilla TOS. I'd like the desktops to be the same.

The obvious solution is to use AES4.1 I appreciate it's a bit slow, but I just like it.

Any version of 1.19 I've tried fails with what appears to be a stuck key. Clicking on File->Open shows a Hebrew character (looks like 0xDB?) filling the box rapidly.

This doesn't happen on the 1.18 release, but this, and all versions back to 1.15 fail in a more subtle way: quick repeated clicking of a window arrow or other window widget causes mouse click lockup. The desktop continues running and responding to keypresses, but it's as if the mouse button is permanently down so no further interaction with the window or menu is possible.

Is AES4.1 off the table entirely, is this a genuine bug (the key repeat one looks like it might be) or have I missed something stupid?

All these tests carried out with a vanilla system. /mint folder contains only one line: GEM=c:/aes41.sys.

Cheers,

DH.

### LINKED ISSUE #378: MultiTOS AES does not work with current kernels
MultiTOS AES (the unofficial/unreleased 4.1 from Atari) does not work with current kernels. There's a problem with the keyboard, a key is stuck and there's no way to "unstick" it.

[GEM.SYS.zip](https://github.com/user-attachments/files/16839450/GEM.SYS.zip)

I've tried to both run it directly with GEM=... and launch it from a shell, and with and without NVDI.

mikrosk:
The commit in question assumed that the only use of O_GLOBAL is inside
slip.xif. However this is not true: it was used in both N.AES and AES
4.1 (and perhaps others) as a way of accessing the console device from
AES by every process.

Roughly, it works like this in AES 4.1:

- "MiNT" (the kernel) opens /dev/console for read/write and stores the
  file handle as CON:

- "AESSYS" opens /dev/console for read/write with O_GLOBAL and gets a
  file handle as if it were "MiNT"

- ("AESSYS" somehow distributes returned file handle to everyone
  interested, perhaps as "this is the console device")

- everyone can now read and write from/to that file handle as if they
  were all root processes (i.e. "AESSYS" is treated as "MiNT" and so are
  others like "NEWDESK" or "SCREEN")

Narrow the O_GLOBAL handling only to /dev/console and allow max 32k
opened files per process (currently the limit is set to 1024).

Fixes #143.
Fixes #173.
Fixes #378.

### DIFF
--- a/.scripts/util.sh
+++ b/.scripts/util.sh
@@ -49,6 +49,7 @@ copy_m68k_modules() {
 	cp "$SRC/sys/sockets/xif/asix_${TARGET}.xif" "$SYSDIR/asix.xix"
 	cp "$SRC/sys/sockets/xif/picowifi_${TARGET}.xif" "$SYSDIR/picowifi.xix"
 	cp "$SRC/sys/sockets/xif/plip_${TARGET}.xif" "$SYSDIR/plip.xif"
+	cp "$SRC/sys/sockets/xif/slip_${TARGET}.xif" "$SYSDIR/slip.xif"
 	cp "$SRC/sys/xdd/audio/.compile_$TARGET/audiodev.xdd" "$SYSDIR"
 	cp "$SRC/sys/xdd/flop-raw/.compile_$TARGET/flop_raw.xdd" "$SYSDIR"
 }
--- a/sys/console.c
+++ b/sys/console.c
@@ -316,7 +316,17 @@ sys_c_auxos (void)
 long _cdecl
 sys_f_instat (int fh)
 {
-	PROC *proc = get_curproc();
+	PROC *proc;
+
+# if O_GLOBAL
+	if (fh & 0x8000)
+	{
+		proc = rootproc;
+		fh &= ~0x8000;
+	}
+	else
+# endif
+		proc = get_curproc();
 
 	if (fh < MIN_HANDLE || fh >= proc->p_fd->nfiles)
 	{
@@ -330,7 +340,17 @@ sys_f_instat (int fh)
 long _cdecl
 sys_f_outstat (int fh)
 {
-	PROC *proc = get_curproc();
+	PROC *proc;
+
+# if O_GLOBAL
+	if (fh & 0x8000)
+	{
+		proc = rootproc;
+		fh &= ~0x8000;
+	}
+	else
+# endif
+		proc = get_curproc();
 
 	if (fh < MIN_HANDLE || fh >= proc->p_fd->nfiles)
 	{
@@ -344,7 +364,17 @@ sys_f_outstat (int fh)
 long _cdecl
 sys_f_getchar (int fh, int mode)
 {
-	PROC *proc = get_curproc();
+	PROC *proc;
+
+# if O_GLOBAL
+	if (fh & 0x8000)
+	{
+		proc = rootproc;
+		fh &= ~0x8000;
+	}
+	else
+# endif
+		proc = get_curproc();
 
 	if (fh < MIN_HANDLE || fh >= proc->p_fd->nfiles)
 	{
@@ -358,7 +388,17 @@ sys_f_getchar (int fh, int mode)
 long _cdecl
 sys_f_putchar (int fh, long c, int mode)
 {
-	PROC *proc = get_curproc();
+	PROC *proc;
+
+# if O_GLOBAL
+	if (fh & 0x8000)
+	{
+		proc = rootproc;
+		fh &= ~0x8000;
+	}
+	else
+# endif
+		proc = get_curproc();
 
 	if (fh < MIN_HANDLE || fh >= proc->p_fd->nfiles)
 	{
--- a/sys/dosfile.c
+++ b/sys/dosfile.c
@@ -48,14 +48,34 @@ sys_f_open (const char *name, short mode)
 	struct proc *p = get_curproc();
 	FILEPTR *fp = NULL;
 	short fd = MIN_OPEN - 1;
+	int globl = 0;
 	long ret;
 
 	TRACE (("Fopen(%s, %x)", name, mode));
 
-	if (mode & O_GLOBAL) {
-	   ALERT("O_GLOBAL is obsolete, please update your driver (%s)",name);
-	   return EINVAL;
-        }
+# if O_GLOBAL
+	if (mode & O_GLOBAL)
+	{
+		TRACE (("O_GLOBAL Fopen(%s, %x)", name, mode));
+
+		if (stricmp (name, "u:\\dev\\console") == 0 || stricmp (name, "u:\\pipe\\sld") == 0)
+		{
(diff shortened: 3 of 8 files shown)


mikrosk:
Btw, Helmut's change in ebbc3f33 does not remove the `O_GLOBAL` flag, just ignore it but I think his original proposed patch is cleaner.

mikrosk:
Fixes also #378 & #173.

th-otto:
The `ALERT` should also be removed. Function like Fopen() should just return an error code, not display an alert box on error.

mikrosk:
Detection is easy (comparing `name` to `/dev/console`) however I'm worried about the double-click issue with AES 4.1 -- that seems to need the whole O_GLOBAL deal, not just the workaround. I'll take a deeper look.

mikrosk:
Yeah, it's because AES 4.1 calls `sys_f_instat` heavily, probably missing some events in the process. I have posted my idea to freemint-discuss for further approval.

mikrosk:
I have updated the branch with a brand new implementation (and updated the description in this PR). It's basically the simplest way. Forcing some specific file handle values to be reserved for `/dev/console` doesn't seem so good after all as this would limit the number of processes which are allowed to open `/dev/console` directly to one.

mikrosk:
I tried to implement @th-otto's suggestion of using FILEPTR->flags. Unfortunately, it wont work, same as my idea of having a dedicated handle wont work. Even if we assume that only one process (AESSYS) opens /dev/console.

The main issue is that we still need some kind of mapping.

If we let /dev/console to be opened as the root process ("MiNT"), we need to map the root process' file handle back to any process accessing /dev/console using that handle (i.e. subtracting 100 from the handle value as done in the original code).

If we let /dev/console to be opened as usual (i.e. by "AESSYS"), we need to map *that* process' file handle back to any *other* process which is accessing /dev/console using that handle. So not only we need to store the handle value, we also need to store the process ID of "AESSYS" which opened /dev/console.

In neither case we can use a flag in FILEPTR, because we get the FILEPTR pointer using an index represented by the file handle. So the file handle must already contain some hint _which_ FILEPTR should be used.

So I propose to merge this PR as is, it really is the simplest solution and it works well even with multiple Fopen("/dev/console") calls. It could even support SLIP.XIF in the same fashion. And it has been verified as working by another user in https://github.com/freemint/freemint/issues/173#issuecomment-2355164031.

th-otto:
On Samstag, 21. September 2024 16:16:36 CEST Miro Kropáček wrote:
> In neither case we can use a flag in FILEPTR, because we get the FILEPTR
> pointer using an index represented by the file handle

Shouldn't something like this work?

In fp_get(), adding something like

if ((*fp)->flags & O_GLOBAL)
   *fp = rootproc->console_handle

(and similar in fp_get1)

Then in sys_f_open, make an entry of the FILEPTR both to the process file 
handle table, mark that with O_GLOBAL, and store that also to rootproc-
>console_handle.


That should work atleast for /dev/console (but not for others since pre-
process handles may be different)

mikrosk:
> Shouldn't something like this work?
>
> if ((*fp)->flags & O_GLOBAL)
>   *fp = rootproc->console_handle

No, because `fp` is uninitialised at this point, you can't use it for the check against O_GLOBAL. Its initialisation is a bit hidden a few lines below: `!(*fp = (*p)->p_fd->ofiles[*fd])`.

th-otto:
Yes, i know, i meant to add those line at the *end* of the function.

mikrosk:
> Yes, i know, i meant to add those line at the end of the function.

Wouldn't this corrupt some internal structures related to /dev/console? I mean, the first user of /dev/console is the root process which opens it and stores as file handle 0. Then AESSYS opens it, too (with O_GLOBAL) and in the original implementation it gets _another_ root process handle (usually 6). Now AESSYS would get 0. So what if those two processes read from / write to /dev/console in parallel? I know it's not a regular file but still, some internal per-file and per-process values may get confused?

mikrosk:
Actually, I had a similar code done and now I remembered why it didn't work: console's functions like `sys_f_instat` and `sys_f_getchar` do not rely on previous Fopen("/dev/console"), that's the whole point of O_GLOBAL. So AESSYS opens /dev/console ... gets file handle 8 ... and then later, process "NEWDESK" comes and starts accessing stuff via that file handle = 8 which is of course bogus.

th-otto:
Oh, ok. So a *different* process uses the same handle, that won't  work of course with my approach. I just wonder: how is the handle passed from N.AES to NEWDESK? Just by keeping it open when AES executes it? If that is the case, then the approach could still work. You just have to adopt the code that inherits the handles from the parent.

mikrosk:
>  I just wonder: how is the handle passed from N.AES to NEWDESK? Just by keeping it open when AES executes it? If that is the case, then the approach could still work. You just have to adopt the code that inherits the handles from the parent.

Frankly, I don't know. I mentioned it in the first post, AES (I'm testing the 4.1 version by Atari, not N.AES) somehow tells everyone ("SCREEN", "NEWDESK", ...) which handle they are supposed to use for accessing the console. And then they do, bypassing Fopen("/dev/console") of their own. Why they can't just open /dev/console? No clue...

mikrosk:
I couldn't come up with anything smarter and since it fixes the problem, I'm merging it. In the future we can rework this again, now we are at least aware of all the side effects and bumps along the round.

Oh and as a bonus, we get slip.xif again. Not that I expect anyone testing it anytime soon but it's there, usable for anyone interested.

th-otto:
I'm not really happy with it, since it is an ugly hack. It could even imagine that some people now think they can use O_GLOBAL again.

>Oh and as a bonus, we get slip.xif again.

Not really, since it is not even compiled currently.

mikrosk:
Thorsten, sometimes I really wish you verified facts before writing your replies.

1. The code is limited to opening "u:\dev\console" and "u:\pipe\sld" as this is the only use case which makes sense for O_GLOBAL. I'm not sure who else could make some use of those.

2. This PR enabled slip.xif building, it's already in the snapshots.

czietz:
If someone is very keen to test, this device uses SLIP: https://www.chzsoft.de/site/hardware/connecting-the-atari-st-to-the-internet/.

(But I have ripped the hardware apart to use it in other projects; so I won't test.)

PR #389, opened 2024-11-06
Adding Raven specific support

agranlund:
Some initial support for Raven Atari clone.

- recognise it as a machine type
- skip SFP detection since it cannot bus error on the tested address and 060FPU+SFP in the _FPU cookie will confuse certain programs
- page0 protection in the same way as Hades and Milan
- mfp.xdd can work on this machine

### DIFF
--- a/sys/arch/info_mach.c
+++ b/sys/arch/info_mach.c
@@ -79,6 +79,9 @@ char *machine_str (void)
 		case machine_ct60:
 			str = "Atari Falcon/CT60";
 			break;
+		case machine_raven:
+			str = "Raven";
+			break;
 # ifdef WITH_NATIVE_FEATURES
 		case machine_emulator:
 			str = nf_name();
--- a/sys/arch/init_mach.c
+++ b/sys/arch/init_mach.c
@@ -57,7 +57,8 @@ enum special_hw
 	none = 0,
 	hades,
 	ct2,
-	ct60
+	ct60,
+	raven
 # ifdef WITH_NATIVE_FEATURES
 	,
 	emulator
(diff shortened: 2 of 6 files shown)


vinriviere:
> Just a question: if the raven does not bus error on the SFP address, how do Pure-C compiled programs work? Their startup code also does this check, and if it does not bus-error, it is assumed to be present, and certain FPU function will try to use it.

Interesting. This means that such Pure-C programs will also have trouble on non-Atari hardware (i.e. EmuTOS/FreeMiNT for Amiga).

skaftetryne:
I think it will only affect PureC programs linked with the 881 library, which I don't think are that common. The _fpuinit symbol is present in both pcstdlib and pc881lib, if pc881lib is not linked (or after pcstdlib in the project file) the code in pcstdlib will be called and I suspect that it does not actually check anything.

th-otto:
Only those that actually use math function of course (the startup code does the check nevertheless). 

Hard to tell which programs are actually affected (Texel might be such a candidate, although it is compiled by Pure-Pascal, it has a similar runtime library).

But of course it would be easy to hack up some simple test program.

th-otto:
> I think it will only affect PureC programs linked with the 881 library

No, definitely not. That will use the real FPU of a 030, not the SFP004.

agranlund:
> Just a question: if the raven does not bus error on the SFP address, how do Pure-C compiled programs work? Their startup code also does this check, and if it does not bus-error, it is assumed to be present, and certain FPU function will try to use it.

I don't know how PureC checks for SFP, but if it does it by bus error like you explained then those programs would behave badly and I suppose you'd might want to use FPUPATCH to patch the program to skip the SFP test and fall back of soft float.
If the program was built for SFP rather than 6888x of course, so perhaps mostly older pre-TT and Falcon things?
https://www.atari-forum.com/viewtopic.php?t=40575

> And another question: who is setting the RAVN cookie?
Ravens TOS extensions.
Currently a mandatory auto program for dev convenience will end up in ROM same as CT60.

PR #391, opened 2024-12-02
Fixed FPU info bug when both FPU + SFP exist

agranlund:
Xaaes system info reported "None" for FPU on systems that has both an FPU + SFP.
The switch statement was not consider that the fputype variable contains the SFP bit, making the whole switch statement invalid.


mikrosk:
Nice catch, thanks!

PR #401, opened 2025-01-17
Added _SC defines + _SC_AVPHYS_PAGES & fix _SC_PHYS_PAGES

MedourMehdi:
I don't know where would be the best place to put _SC defines?

### DIFF
--- a/sys/dos.c
+++ b/sys/dos.c
@@ -578,7 +578,7 @@ sys_s_ysconf (int which)
 {
 	switch (which)
 	{
-		case -1:	return 9;
+		case -1:	return 10;
 		case  0:	return UNLIMITED;
 		case  1:	return 32767; /* matches ARG_MAX */
 		case  2:	return NDFILE;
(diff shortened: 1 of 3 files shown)


mikrosk:
If I may suggest one thing, try to split this into two commits:

- adding totalphysmem()
- rename (and add a comment about the need to sync this with `mintlib/include/bits/confname.h`)

Also, perhaps keeping the order of numbers is better than grouping PHYS/AVPHYS together so someone doesn't overlook it when adding number 11.

MedourMehdi:
I'm not an expert of gitology. Is it ok now?

mikrosk:
Force-push is the way to go, yes. But what I meant is that you add totalphysmem() and `case 10` and then, in the other commit, you refactor it (i.e. add names to the numbers). That way it will be more obvious what functionality has been added and what was just a cleanup commit.

th-otto:
I think so, but @mikrosk : is there some way to display a squashed version of the commits?

mikrosk:
I'm always using the "Files changed" tab to see the squashed version but in this case I'd like to merge it as rebase, i.e. as separate commits.

MedourMehdi:
Thanks @th-otto, should be fixed now.

PR #410, opened 2025-05-30
fix(net/ppp): add safe_memcpy for unaligned buffer handling

MedourMehdi:
- Added safe_memcpy implementation to handle unaligned buffers
- Replaced standard memcpy with safe_memcpy in ppp_recv_frame
- Implemented direction-aware copying to prevent data corruption
- Ensures proper handling of overlapping memory regions

### DIFF
--- a/sys/sockets/xif/ppp.c
+++ b/sys/sockets/xif/ppp.c
@@ -187,6 +187,29 @@ static ushort const fcstab[256] =
 # define FCS_INIT	0xffffu
 # define FCS_GOOD	0xf0b8u
 
+#ifndef __SIZE_T
+#define __SIZE_T
+typedef unsigned long size_t;
+#endif
+
+static void safe_memcpy(void *dest, const void *src, size_t n)
+{
+    char *d = (char *)dest;
+    const char *s = (const char *)src;
+    if (d < s) {
+        while (n--) {
+            *d++ = *s++;
+        }
+    } else {
+        d += n;
+        s += n;
+        while (n--) {
+            *--d = *--s;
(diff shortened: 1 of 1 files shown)


th-otto:
??? All the cases you mention (unaligned addresses, overlapping regions) are already handled by the standard implementation of memcpy.

MedourMehdi:
Hi the problem was that it won't build with gcc > 7.5 - at least this is my workaround in order to make the kernel build with gcc 9.3 and highter.

MedourMehdi:
I don't know how to silence it. I have

```
m68k-atari-mint-gcc -I.. -I../.. -D__KERNEL_XDD__  -m68000   -fgnu89-inline -Werror -mshort -g -O2 -fomit-frame-pointer -fno-strict-aliasing -Wall -Wmissing-prototypes -Wshadow -Wpointer-arith -Waggregate-return  -Wp,-MD,.deps/ppp_000.P -Wp,-MT,ppp_000.o -c ppp.c -o ppp_000.o
In function 'ppp_recv_frame',
    inlined from 'ppp_recv' at ppp.c:480:5:
ppp.c:865:3: error: 'memcpy' accessing between 3 and 2147483646 bytes at offsets -1 and 0 overlaps between 2 and 2147483645 bytes at offset 0 [-Werror=restrict]
  865 |   memcpy (cp-1, cp, len);
  ```

th-otto:
Done now. Should make this PR obsolete.

MedourMehdi:
Cool,
Many thanks Thorsten.

PR #420, opened 2025-08-23
Update detect.S

Landemarre:
68080 use 68060 stack frame format so need to be managed as 68060 not 68040

### DIFF
--- a/sys/arch/detect.S
+++ b/sys/arch/detect.S
@@ -313,7 +313,6 @@ x040:	moveq	#40,d0		// assume 68040
 	dc.l	0x06d00000	// check for 68080 addiw.l instruction
 
 // Apollo 68080
-	moveq	#40,d0          // force 68040 detection
 	move.w	#1,SYM(is_apollo_68080)
 	bra.s	exit
 


vinriviere:
Long time ago, I asked Gunnar which Motorola CPU the Apollo 68080 was supposed to be compatible. He answered 68040, so I did. Does the situation have changed? Of course, if the Apollo 68080 produces a 68060 stack frame, this was wrong. Which stack frames are different? Which real-world trouble this was causing?

th-otto:
IIRC, the only frames that are different between 040 and 060 are the ones for bus- and address errors.

Whats more important is maybe: does the apollo really implement all the bits of the 060 CACR, like the ones for branch prediction cache?

Landemarre:
> IIRC, the only frames that are different between 040 and 060 are the ones for bus- and address errors.
> 
> Whats more important is maybe: does the apollo really implement all the bits of the 060 CACR, like the ones for branch prediction cache?

Apollo not support CACR bit for branch prediction cache, as it not work as 68060 perhaps just not need.
New MMU look will be fully compatible with 68060

PR #427, opened 2025-10-04
Small cleanup in context/interrupt handlers

mikrosk:
Saving SR just before accessing MMU regs in save_context and change_context has no real purpose.

The comment "at IPL <level>" means two different things in mint_vbl. Once it is "running at IPL 6" (i.e. the IPL mask = 5 or lower) and once it is "the IPL mask = 3 or lower" (therefore running at IPL 4). So change "at IPL 3" to "at IPL 4" to avoid confusion.


mikrosk:
One could also argue whether
```
move.w	#0x2400,sr		// lower the current IPL
```
is exactly descriptive as it is actually increasing the IPL value from 0..3 to 4.

PR #432, opened 2025-11-17
Check for special Fforce() handles

### LINKED ISSUE #431: sys/k_fds.c:fp_get(): O_GLOBAL change breaks use of negative file descriptors (e.g., Fforce(2, -1))
When mintlib starts up a C program, it attempts to force stderr to the console using  Fforce(2,-1) [here](https://github.com/freemint/mintlib/blob/fba33c97fb2979edc0f9133ca54b133e1a66d707/mintlib/main.c#L164).

But this is rejected by `sys_f_force()` with EBADF. From a debug log:

```
pid   1 (dash): sys_f_force(): invalid fd handle (32767)!
```

The `fp_get()` function in sys/k_fds.c thinks the file descriptor is 32767 (not -1) because the O_GLOBAL related code there has reset the high bit.

--
I encountered this when attempting to run the dash shell via INIT=/bin/dash. I saw that the prompt and error messages from dash were sent to the serial port, indicating that Mintlibs's attempt to redirect stderr to the console wasn't working.

mikrosk:
Fixes #431.

@codesmythe can you build your own kernel and check whether the issue has been resolved for you?

### DIFF
--- a/sys/k_fds.c
+++ b/sys/k_fds.c
@@ -150,7 +150,7 @@ long
 fp_get (struct proc **p, short *fd, FILEPTR **fp, const char *func)
 {
 # if O_GLOBAL
-	if (*fd & 0x8000)
+	if ((*fd < -5 || *fd > -1) && (*fd & 0x8000))
 	{
 		*fd &= ~0x8000;
 		*p = rootproc;


codesmythe:
Yes, I can confirm that this resolves the issue I was seeing. Thanks for the fix!

PR #437, opened 2026-02-19
Work to get slattach & slip.xif working again. O_GLOBAL gatekeeping removed.

dh219:
As per conversation on the mailing list, O_GLOBAL is used in slip.xif to open the serial device. In theory the gatekeeping could be expanded to the /dev directory, but with the mitigations to the number of open files caused by the move to use bit 7 as the flag, I think it less confusing to remove it entirely and restore O_GLOBAL to be available to any file.

This patch does that, sets the bit pattern used to identify a global as a #define, removes a couple of broken debug lines and changes slattach from referencing /sbin to /bin.

The latter I may try to rework further, but for now this is consistent with the snapshots produced and is tested working on a real system and Hatari.

Builds on #379

### DIFF
--- a/sys/console.c
+++ b/sys/console.c
@@ -319,10 +319,10 @@ sys_f_instat (int fh)
 	PROC *proc;
 
 # if O_GLOBAL
-	if (fh & 0x8000)
+	if (fh & O_GLOBAL_BITPATTERN)
 	{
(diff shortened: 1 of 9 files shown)


mikrosk:
No longer necessary but I took the slip & slattach bits.

PR #449, opened 2026-06-26
Introduce graceful shutdown to kernel modules (xdd, xif, km)

mikrosk:
I have reworked this branch at least three times from scratch and I'm still not sure whether this is the best approach. :-)

We really suffer with having so many different kinds of kernel modules. The gist of this change is described in the first commit, TL;DR version is that Svethlana uses a non-standard exception vector (same as EtherNAT and Lightning does) so after a soft reset its interrupts are still active and crashing CT60's memory test.

And while doing so, I have noticed that USB drivers try to work around this exact issue but as far as my testing goes, badly: the reset vector is honoured *after* the CT60 memory test, therefore hooking on it to disable Svethlana's interrupt wasn't enough. So I reworked that, too.

@DavidGZ @czietz @claude as you have been the most active contributors to the USB part, feel free to raise your concerns. Especially about the 3rd commit (removal of the low-level ioctl calls).

In the future I'd like to rework at least the *.km part but I'm afraid *.xif are stuck with their API forever (there is no versioning and quite a few closed-source drivers, Ozk's being the most prominent ones). I had implemented an ioctl() API similar to the USB oen, too but then I realised that it is perhaps over-engineered (hence my final 3rd commit also for USB).

### DIFF
--- a/sys/sockets/inet4/if.h
+++ b/sys/sockets/inet4/if.h
@@ -265,5 +265,6 @@ long		if_open		(struct netif *);
 long		if_close	(struct netif *);
 long		if_send		(struct netif *, BUF *, ulong, short);
 void		if_load		(void);
+void		if_stop		(void);
 
 # endif /* _if_h */
--- a/sys/biosfs.c
+++ b/sys/biosfs.c
@@ -367,6 +367,32 @@ set_xattr (XATTR *xp, ushort mode, int rdev)
 
 
 void
+shutdown_all_devices (void)
+{
+	struct bios_file *b;
+
+	for (b = broot; b; b = b->next)
+	{
+		struct bios_file *p;
+
+		if (b->drvsize <= (long) offsetof (DEVDRV, shutdown)
+		    || !b->device || !b->device->shutdown)
+			continue;
+
+		/* A driver may install several u:\dev nodes that share one
+		 * DEVDRV (e.g. the two SCC ports); call its shutdown() once.
+		 */
+		for (p = broot; p != b; p = p->next)
+			if (p->device == b->device
+			    && p->drvsize > (long) offsetof (DEVDRV, shutdown))
+				break;
+
+		if (p == b)
+			(*b->device->shutdown) ();
+	}
+}
+
+void
 biosfs_init (void)
 {
 	struct bios_file *b, *c;
--- a/sys/biosfs.h
+++ b/sys/biosfs.h
@@ -69,6 +69,7 @@ extern struct tty sccb_tty, scca_tty, ttmfp_tty;
 extern long rsvf;
 
 void biosfs_init (void);
+void shutdown_all_devices (void);
 
 long rsvf_ioctl (int f, void *arg, int mode);
 long iwrite	(int bdev, const char *buf, long bytes, int ndelay, struct bios_file *b);
--- a/sys/dos.c
+++ b/sys/dos.c
@@ -25,11 +25,13 @@
 # include "mint/signal.h"
 # include "mint/stat.h"
 
+# include "biosfs.h"		/* shutdown_all_devices */
 # include "console.h"
 # include "cookie.h"
 # include "filesys.h"
 # include "k_prot.h"
 # include "memory.h"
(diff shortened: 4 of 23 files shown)


DavidGZ:
I don't have a strong opinion about removing the low-level init/stop functions. I think it all comes down to what we expect MiNT to become in the future.

If we realistically think that MiNT is not going to undergo major changes or gain significant new features, and that most future work will consist of bug fixes, then I think we can safely remove them to simplify the code.

On the other hand, if we expect MiNT to continue evolving and adding new features, those lowlevel_init/lowlevel_stop functions could be useful. For example, they could support a scenario where the module remains loaded but the hardware is temporarily stopped, such as for some form of power management.

sys/sockets/inet4/if.h
```diff
@@ -265,5 +265,6 @@ long		if_open		(struct netif *);
 long		if_close	(struct netif *);
 long		if_send		(struct netif *, BUF *, ulong, short);
 void		if_load		(void);
+void		xif_stop	(void);
```
DavidGZ: A minor note:

Why did you change the function prefix? As fa as I can see, the rest of the functions there have "if_*" as prefix rather than  "xif_*"

czietz:
Personally, I'm not a fan of USB API changes; since it breaks third-party applications: for example, uip tool, the STiNG driver for USB, etc. It makes it necessary to provide two versions of these apps (for old and new API), and that really confuses users who then tend to blame the app instead of the FreeMiNT USB API change.

sys/sockets/inet4/if.h
mikrosk: The idea was to make it similar to usb_stop(), i.e. to emphasise that it is the _driver_ shutdown, not an interface one (hence the void argument). But I agree it looks very out of place.

mikrosk: Hm, now I see that `if_load`, just one line above, literally just loads all *.xif modules. so `if_shutdown` (which shuts down, again, all modules) actually isn't so mislabeled.

mikrosk:
Good point, @czietz. There's no other gain than parity with xif open/close. Those do all the work, i.e. there is no need for a shutdown, so I was wondering what ucd open/close do and well, they do nothing, since their inception.

Taking also @DavidGZ's input into consideration, I'm going to drop the 3rd commit and leave the open/close/init/deinit as it was.

mikrosk:
@DavidGZ I couldn't resist one "parity change" while here: usb_stop() used to tear each controller down via ucd_unregister(), which I don't think is right for a shutdown hook:
  
  - ucd_unregister() is the full unload operation: it closes the controller, issues LOWLEVEL_STOP, and unlinks the ucdif from `allucdifs`. In the shutdown path the only thing that matters is shutting the hardware down (so a stray IRQ doesn't fire into the memory test) -- the list is discarded with the system a moment later anyway, so unlinking from it is pointless.
  - As a side effect it also mutates `allucdifs` while usb_stop() is iterating it. That happens to be safe today only because ucd_unregister() leaves `a->next` intact (it relinks the predecessor, never NULLs the removed node) — an unguaranteed implementation detail. If someone ever "tidies up" ucd_unregister() to also clear `a->next`, the loop would terminate early and leave controllers running.

So usb_stop() now just calls close() + LOWLEVEL_STOP on each controller, which also mirrors what if_stop() does on the inet4 side (nif->close(), no deregister). The parity isn't complete though: inet4 has a separate if_open()/if_close() layer (driven by ifconfig up/down) on top of the nif->open/nif->close vectors, whereas USB has no ucd_open()/ucd_close()  - a controller is opened in ucd_register() and closed only in ucd_unregister(). So usb_stop(), like if_stop(), goes straight to the driver's close() vector.

As an interesting side effect, ucd_unregister() now isn't called from anywhere (while still being available via usb_api.ucd_unregister of course) - same as if_deregister() on the inet4 side, which is exported for drivers but has no in-tree caller either.

DavidGZ:
@mikrosk OK. Thanks for implementing this!

